PR #17, opened 2019-02-01
fix(logger): remove logger truncation

Birkbjo:
See https://jira.dhis2.org/browse/DHIS2-5853 comments for reasons for removing truncation. 
Not that there was a bug that made the client create the timestamp for summaries. This would make the timestamp be different from the server (like 1 hr after the other messages). I've removed the 'fake information' and replaced it with a timestamp, as according to Joe's comment. 

I decided to not remove the functionality of truncation, but added a flag for this. Mostly due the not knowing the reasons for doing this in the first place. I could've probably simplified it and removed the functionality all together, what do you think?

Also upgraded code-style and fixed the tests.
Sorry for upgrading code-style in a bug-fix, but it was the easiest way out, as the previous version staged all files.

### DIFF
--- a/src/components/Logger/index.js
+++ b/src/components/Logger/index.js
@@ -190,12 +190,17 @@ export class Logger extends React.Component {
         )
     }
 
-    getDate(p, prevDateHH, prevDate) {
+    getDate(p, prevDateHH, prevDate, truncate = false) {
         if (!p.d) {
             return null
         }
-
         let date = moment(p.d).format('YYYY-MM-DD HH:mm:ss')
+
+        if (truncate && moment(p.d).format('YYYY-MM-DD HH') === prevDateHH) {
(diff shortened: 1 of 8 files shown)


src/components/Logger/index.js
```diff
@@ -190,12 +190,15 @@ export class Logger extends React.Component {
         )
     }
 
-    getDate(p, prevDateHH, prevDate) {
+    getDate(p, prevDateHH, prevDate, truncate = false) {
```
Mohammer5: Can you change the variable name to something meaningful or add a jsdoc commment?
I think that variables shouldn't be abbreviated with single letters as it's totally unclear what they mean.

Birkbjo: I agree. I just have a habit of not refactoring code I've not written (probably due to my work on maintenance, hehe). Done now.

PR #24, opened 2019-03-12
DHIS2-6061 - add csv checkbox

ismay:
https://jira.dhis2.org/browse/DHIS2-6061


ismay:
Closed. App should be refactored before we add new features.

PR #27, opened 2019-03-21
fix/dhis2 5706/fix data summary

Birkbjo:


src/pages/import/helpers.js
```diff
@@ -145,6 +145,8 @@ export async function fetchTaskSummary(jobId, type) {
                     'summary.importSummaries',
                     data.importSummaries
                 )
+            } else if (data.importCount) {
+                eventEmitter.emit('summary.importCount', data.importCount)
             }
```
varl: Could we put an `else` clause here just to e.g. log an error? It's difficult to debug these cases when the function just returns `undefined`.

PR #30, opened 2019-03-26
fix(metadata export): filter selectable schemas on metadata and exclude schemas

Birkbjo:
See https://jira.dhis2.org/browse/DHIS2-5680.

When loading schemas, it preselected all schemas before filtering in metadata. Now also filters on the EXCLUDE_SCHEMAS list, as a workaround described in the ticket above. 

There seemed to be some performance problems with the schema-list - both with rendering and fetching unecessary. Ive now put the fetched list in the redux-store so it does not fetch each time an user visits the 'Metadata Export'-page.
Also moved 'checked' to be an Object instead of list, to improve rendering (so each item does not need to loop through the checked list - now just a O(1) operation).

### DIFF
--- a/src/components/Form/Schemas/index.js
+++ b/src/components/Form/Schemas/index.js
@@ -9,8 +9,6 @@ import { setSchemas } from 'reducers/'
 import { connect } from 'react-redux'
 import s from './styles.css'
 import { getSortedSchemaGroups, getSchemas } from 'reducers/schemas/selectors'
-import { colors } from 'material-ui/styles'
-import { EditorFormatAlignJustify } from 'material-ui/svg-icons'
 
 function groupName(klass) {
     let group = klass.split('.')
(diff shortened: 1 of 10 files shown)


src/components/Form/Schemas/index.js
```diff
@@ -173,7 +173,7 @@ export default class Schemas extends React.Component {
 
     getSchemas(schemas) {
         return schemas
-            .filter(i => i.metadata)
+            .filter(i => i.metadata && !EXCLUDE_SCHEMAS.has(i.collectionName))
             .map(i => ({
```
HendrikThePendric: This is nice and readable, but slow. If this is executed often, perhaps replace with a single `reduce`

Birkbjo: It's not slower than before. And Schemas has at most like 90 elements. Also why I changed EXCLUDE_SCHEMAS to be a set instead of an array, so we dont need a O(n) lookup for every element. 
It's also only run once (per metadata-export page visit); when the Schemas component mounts. I will actually look into caching the schemas, as its unecessary to load this every time you visit the page.

PR #32, opened 2019-03-27
Dhis2 5292/skip sharing metadata dep export

Birkbjo:
Most of the code is refactoring for generating download URL. To clean up the OnSubmit functions.

### DIFF
--- a/src/helpers/download.js
+++ b/src/helpers/download.js
@@ -1,3 +1,5 @@
+import { api } from 'services'
+
 export function blobType(format, compression) {
     if (compression === 'gzip') {
         return `application/${format}+gzip`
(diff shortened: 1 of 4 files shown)


src/pages/export/MetaData.js
```diff
@@ -69,30 +69,21 @@ export class MetaDataExport extends FormBase {
...
-            }
+            let endpoint = `metadata`
+            const downloadUrl = getDownloadUrl({
+                format,
+                compression,
+                endpoint,
+                sharing,
+            })
+            const schemaParams = schemas
+                .map(name => name)
```
HendrikThePendric: I think this map can be removed

src/helpers/download.js
```diff
@@ -25,3 +27,12 @@ export function downloadBlob(url, filename) {
     document.body.appendChild(link)
     link.click()
```
HendrikThePendric: should we not clean up after ourselves and remove the link from the DOM?

src/pages/export/MetaData.js
```diff
@@ -69,30 +69,21 @@ export class MetaDataExport extends FormBase {
...
+                .sort()
+                .map(name => `${name}=true`)
+                .join('&')
 
-            let endpoint = `metadata${format}${
-                compression !== 'none' ? compression : ''
-            }`
-            window.location = api.url(`${endpoint}?${params.join('&')}`)
+            const url = downloadUrl.concat(`&${schemaParams}`)
```
HendrikThePendric: `downloadUrl` is a string, I would find it clearer if you used a template string here instead of concat.

PR #35, opened 2019-05-23
chore: uneject

ismay:
Check during review

- package name was missing a dash, fixed it, I assume that it's ok to change the package name?
- moved to unversioned endpoints (see files where `REACT_APP_DHIS2_BASE_URL` is used)

Done

- [x] fix absolute imports
- [x] fix css modules
- [x] `__tests__` folders
- [x] add back used dependencies
- [x] apply hosting on nested paths fix
- [x] fix package.json scripts
- [x] check usage of decorators
- [x] check `{ exports }`
- [x] fix deep imports that can be more shallow
- [x] `mount` undefined in tests
- [x] reenable travis
- [x] fix tests
- [x] scripts
- [x] config
- [x] i18n
- [x] src
- [x] public
- [x] fix manifest.json
- [x] package.json
- [x] cli-style
- [x] travis config
- [x] readme

Follow up issues

- remove eslint ignore comments (max-params)
- index files do re-exporting as well as define components
- maybe centralize api endpoint config (see ./src/config.js and ./services/API)?
- getting some errors about side effects for unmounted components
- why is it using an evenemitter, and xhr?


Mohammer5:
I can't say anything about ejecting/unejecting, have to read into that.
But I can `yarn start` the app and everything seems to work fine. `yarn build` works as well

ismay:
Basically I just create a new create-react-app with the latest version of create-react-app, and then just put everything back bit by bit. `./config` and `./scripts` can be removed completely as `react-scripts` manages them. Same with a lot of the dependencies.

Besides that there are usually some updates, like css modules now needing a `.module.css` prefix to be imported correctly with react-scripts. And along the way I also try and fix things if I encounter something and it's not too big. If it's a larger change I'll leave it as a follow-up issue.

ismay:
Plus applying cli-style (if it's not already there), and switching to the `REACT_APP_DHIS2_BASE_URL` (instead of discovering the base url through the manifest). Which could be left out if we'd prefer to do them separately.

varl:
Looks good to me. Really like this checklist you got going on for unejecting apps. :+1:

ismay:
> Looks good to me. Really like this checklist you got going on for unejecting apps.

Thanks! Yeah helps keep track of everything ✅✅✅

PR #43, opened 2019-06-28
feat: add firstRowIsHeader param for csv meta data import

Mohammer5:
Fixes: [DHIS2-6061](https://jira.dhis2.org/browse/DHIS2-6061)

While adding the parameter I found the following bug which should be considered when reviewing this PR: [DHIS2-7221](https://jira.dhis2.org/browse/DHIS2-7221)

### DIFF
--- a/src/helpers/fields.js
+++ b/src/helpers/fields.js
@@ -55,6 +55,12 @@ const fields = {
         i18n.t('Event ID Scheme'),
         TYPE_RADIO
     ),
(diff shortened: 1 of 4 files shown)


ismay:
Let me preface this by saying that I find this app very difficult to review properly, there are a ton of uncommon patterns. I've tested with the play 2.32.0 backend though, and the csv import works as expected for me.

PR #45, opened 2019-07-09
refactor(import meta data): rename import mode labels to dry run

Mohammer5:
Fixes [DHIS2-6379](https://jira.dhis2.org/browse/DHIS2-6379)

Import mode is used in meta data import only, so simply renaming it should be safe.
I've moved it up to where the old "Dry run" previously was.


ismay:
Makes sense. Though it is a bit confusing, but this part of the app is to blame for that. Only way to make it less confusing would be if we refactor this entire part of the app where the fields get rendered, I think.

So tldr, lgtm 👍

PR #50, opened 2019-07-12
fix: use selected format instead of file extension for Content-Type header

Mohammer5:
This PR will change the way what Content-Type header is being sent with the request.
This will allow to send .zip files without having to have .json.zip as file extension.

Fixes DHIS2-6247 (https://jira.dhis2.org/browse/DHIS2-6247)

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2019-04-09T14:14:19.008Z\n"
-"PO-Revision-Date: 2019-04-09T14:14:19.008Z\n"
+"POT-Creation-Date: 2019-07-12T06:06:14.220Z\n"
+"PO-Revision-Date: 2019-07-12T06:06:14.220Z\n"
 
 msgid "user is not logged in"
 msgstr ""
@@ -140,9 +140,6 @@ msgstr ""
 msgid "Identifier"
 msgstr ""
 
-msgid "Import Mode"
-msgstr ""
-
 msgid "Report Mode"
 msgstr ""
 
@@ -249,12 +246,6 @@ msgid "Skip check (fast)"
 msgstr ""
 
 msgid "Check (safe, recommended)"
-msgstr ""
-
-msgid "Commit"
-msgstr ""
-
-msgid "Validate"
 msgstr ""
 
 msgid "AUTO"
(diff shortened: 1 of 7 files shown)


HendrikThePendric:
I've carried on testing by just using my local (docker based) dhis2-core instance and running the app in dev mode. Most of this app seems to be broken. And this seems to be the case both on master as well as on this branch. And I have never worked on this app before, so I am not aware of any gotchas or limitations. As such, it is impossible for me to review this PR: I don't know what is supposed to be working and what is supposed to be broken (locally).

Mohammer5:
> When the format (chosen in the UI) and the actual file type don't match, I don't get a warning about this. Instead, I have to read in the collapsed logger what's gone wrong. In the meanwhile, the progress indicator keeps spinning. In the master branch this type of scenario seemed to be handled better. Is there a way you could implement the current changes, but still keep the errors as they were on master?

The current version on master checks the type by extracting it from the file name.
* When the filename contains `json` and you've selected `csv`, it will pick `application/json` as content type. Which is correct and wrong (it shouldn't use the json in case the user picked the wrong file accidentally), anyways: In this case there's no error on the master branch
 * When the filename doesn't contain a file type, the app says it's `null` and therefore shows an error. My changes always send a content type and it's the backend's responsibility to determine whether the file type was correct or not
* When the filetype is not correct, the backend responds with 500 and this branch (for the master) branch displays the following error: `Maximum column count of 100,000 exceeded in record 0. Set the SafetySwitch property to false if you're expecting more than 100,000 columns per record to avoid this error.`

So I'm not sure what you saw, can you give me a description how to reproduce? Ideally with `REACT_APP_DHIS2_BASE_URL=https://play.dhis2.org/dev yarn start`

> I don't think this has anything to do with your changes, because I get this on master too, but when an import is done, the progress bar just keeps going....

https://jira.dhis2.org/browse/DHIS2-7221

> Most of this app seems to be broken.

I don't know what you mean by this? Do you see the same behavior on the play versions? Or just the local version? (To find the import export app on play.dhis2.org/dev, you'll need to log in with the system user, not admin).
@ismay Could this be related to un-ejecting? Because [DHIS2-7221](https://jira.dhis2.org/browse/DHIS2-7221) also seems to have introduced around the time of un-ejecting.

ismay:
> @ismay Could this be related to un-ejecting? Because DHIS2-7221 also seems to have introduced around the time of un-ejecting.

I don't suspect so. I tried to touch as little as possible of the app's internals, so there shouldn't be any significant changes other than bundling etc. I did notice however, during this PR and before, that the app seems to be a lot more complicated than its functionality would warrant. I can imagine bugs crop up in this app because the logic is tricky to follow.

PR #56, opened 2019-07-26
refactor: use react-final-form instead of inheritance pattern

Mohammer5:
With this PR the pages no longer uses the inheritance pattern to build the form and the view in general.

I've used react-final-form to build re-usable components, which we should be able to use in other sections without complications as well

#### Import
* [ ] Meta data
* [x] Data
* [ ] Event
* [ ] GML

#### Export
* [ ] Meta data
* [ ] Event
* [ ] Data
* [ ] Meta data dependency

<hr />

### To Do for data import:

* [x] On selecting a file and clicking the file select again, I get a spinner. That's a bit jarring, and doesn't seem to happen in the version that's on play.
* [x] I like snapshot tests for presentational components, it's useful to be notified that it changed with a diff of the changes. But for testing any component behaviour I prefer asserting that it behaves as expected. So with for example the MoreOptions component I'd prefer using enzyme to test if the elements are/aren't there. Same for click handlers, etc. Otherwise you have to manually scan the snapshot and that's easy to get wrong.
* [x] I'd go with a unitless line-height. Seems simpler to me to have it just be proportional.
* [x] The more options button doesn't have a cursor: pointer.
* [x] Is the negative margin necessary on the more options button? It's tiny, but just using the right margin looks fine to me and seems less magic-numbery.
* [x] Maybe you could lift up the definition of the react-final-form mock to a common __mocks__ folder.
* [x] I know we do this a lot in our apps, but I'd remove the console.logging in production.
* [x] Maybe formfooter and formheader instead of formfoot and formhead?


ismay:
I didn't mention it above, but your changes seem like a big improvement to me. Way easier already to see what's going on! 👍

Mohammer5:
> After uploading a file it's not clear to me how to remove it. Clicking it allows you to upload again but that's not immediately clear to me as a user.

What do you mean with removing it?
You're not uploading a file to store it, you're uploading it to import the contents of said file into the database, after that the file will be dumped

> The uppercase labels next to the radio buttons are quite shouty. I'd prefer sentence case, except for the file extensions:

Good catch, at first I thought the original labels are all uppercase and added text-transform uppercase. I'd like to keep the ones that are uppercase already on master, to make it a plain refactor instead of refactor + improvement (improvement should be a separate PR / commit)

> Clicking a radio button produces a weirdly offset outline:

Noticed that as well, I think it's a `ui-core` issue

> Personally I'd prefer em and rem for the sizing in the css-modules. It seems less magic-numbery to me and I find it easier to create responsive components that way since everything will just inherit from the font-size and you can then just scale that all at once.

I used pixels because the rest of the app is using pixels also. I felt that using rem is an improvement I want to do separately.

> I was wondering where you'd found the signature for the onSubmit callback you're using with final-form. 

It's a higher order function that I use to create a function that follows the signature you found, which accepts `values` as the first arg

> It'd be nice to have all the new components tested as well, though I assume you're working on that since you encountered the error with jest.

I haven't added any tests yet, but surely, I will!

> I also find the nesting of components beyond the src/components/ComponentName folder a bit cluttery. 

I 100% agree, just kept the current style. I could do that in a separate commit as one of the last things before changing this PR from draft to a proper one

ismay:
> What do you mean with removing it?
You're not uploading a file to store it, you're uploading it to import the contents of said file into the database, after that the file will be dumped

Yeah no I don't mean removing it after uploading. I should've said replace, that's more what I mean. Let's say that you selected the wrong file by accident, but haven't submitted yet, so you want to select a different file. It's not that difficult to figure out, but something like a 'Upload a different file' button would be more explicit. But that's probably something for a different PR.

> improvement should be a separate PR / commit

Definitely. Sounds good

> I used pixels because the rest of the app is using pixels also. I felt that using rem is an improvement I want to do separately.

Yup makes sense.

> It's a higher order function that I use to create a function that follows the signature you found, which accepts values as the first arg

Yeah but it doesn't seem to give final-form any feedback on the state of the submission (i.e. success or failure). That way final-form will internally have the form as 'pending' even though submission might have already finished. I'm referring to using one of the three options listed for the onSubmit handler:

> Synchronously: returns undefined on success, or an Object of submission errors on failure
> Asynchronously with a callback: returns undefined, calls callback() with no arguments on success, or with an Object of submission errors on failure.
> Asynchronously with a Promise: returns a Promise<?Object> that resolves with no value on success or resolves with an Object of submission errors on failure. The reason it resolves with errors is to leave rejection for when there is a server or communications error.

Or am I misreading what you're doing here?

Mohammer5:
> Yeah but it doesn't seem to give final-form any feedback on the state of the submission (i.e. success or failure). That way final-form will internally have the form as 'pending' even though submission might have already finished. 

That's not needed in this case, because once loading starts, the form will be unmounted and the `Loading` component will be displayed:
https://github.com/dhis2/import-export-app/blob/8a369f75a508dbe8d3b6ac507dff25964343ea31/src/pages/import/Data.js#L33

I haven't changed that as I didn't want to mess with anything (e. g. the data submittion) but the view.

ismay:
Ah ok that's why. Yeah that makes sense. Should have looked beyond the onSubmit handler 😋

Mohammer5:
@ismay this should be ready to review.
I won't set it to ready for review though as I have another branch that I would like to merge into this one in a separate PR (the meta data import section)

package.json
```diff
@@ -40,7 +43,8 @@
     "@dhis2/cli-style": "^3.2.2",
     "enzyme": "^3.9.0",
     "enzyme-adapter-react-16": "^1.13.2",
-    "husky": "^2.3.0"
+    "husky": "^2.3.0",
+    "react-test-renderer": "^16.8.6"
```
ismay: Personally I think I'd go with enzyme-to-json for your snapshot tests, since enzyme is already part of the codebase. Plus that way you can choose to do shallow rendering when necessary.

src/pages/import/Data/helper.js
```diff
@@ -0,0 +1,111 @@
...
+            () => setLoading(false),
+            setError,
+            format
+        )
+
+        xhr.send(upload)
+    } catch (e) {
+        console.error(e)
+
+        setError({
```
ismay: I think I'd prefer just passing a message instead of this object. So setErrorMessage and errorMessage. Or if you prefer (if you need the stacktrace), passing a normal error. From looking at your changes it doesn't seem like you're obliged to use this specific object right? Or am I missing something?

ismay: You would of course have to change the other parts that deal with these errors, but they're all under your control right?

src/helpers/useErrorHandler.js
```diff
@@ -0,0 +1,16 @@
+import { useState, useMemo } from 'react'
+
+export const useErrorHandler = () => {
+    const [error, setError] = useState('')
+    const errorHandler = useMemo(
+        () => event => {
+            console.log('event', event)
+            try {
+                setError(event.target.response.message)
+            } catch (e) {}
```
ismay: A comment here explaining why it's ok to swallow this error would be nice.

src/pages/import/Data/helper.js
```diff
@@ -0,0 +1,111 @@
...
+                'skipExistingCheck',
+                'strategy',
+            ],
+            append
+        )
+
+        setLoading(true)
+
+        const { REACT_APP_DHIS2_BASE_URL } = process.env
+        const url = `${REACT_APP_DHIS2_BASE_URL}/api/dataValueSets.json?${params}`
```
ismay: I like using url-join for concatenating parts of a url like here (https://www.npmjs.com/package/url-join). Less chance to end up with an invalid url.

src/pages/import/Data/helper.js
Mohammer5: I copied the existing structure which is used by the `getUploadXHR` function in `src/helpers/xhr.js`

Definitely a change we should keep in mind though, I don't like this at all either

Mohammer5:
> On selecting a file and clicking the file select again, I get a spinner. That's a bit jarring, and doesn't seem to happen in the version that's on play.

@ismay I assume you tested the v32 version? On v33 (master) for me it behaves like the version on play: https://play.dhis2.org/dev/dhis-web-import-export/index.html#/import/data

Just the spinner that's still showing once done (bug on play) is not happening for me locally (which is good)

src/pages/import/Data/helper.js
ismay: > I copied the existing structure which is used by the getUploadXHR function in src/helpers/xhr.js

Ok yeah, that's one for later then 👍

ismay:
> > On selecting a file and clicking the file select again, I get a spinner. That's a bit jarring, and doesn't seem to happen in the version that's on play.
> 
> @ismay I assume you tested the v32 version? On v33 (master) for me it behaves like the version on play: https://play.dhis2.org/dev/dhis-web-import-export/index.html#/import/data
> 
> Just the spinner that's still showing once done (bug on play) is not happening for me locally (which is good)

You also mean the loading bar that shows when you want to replace the loaded file right? (I should have said loading bar instead of spinner). I tested on 2.32, because play 2.33 doesn't have import-export at the moment somehow. If this is also in 2.33 I guess it's not a regression and we can deal with it separately.

Mohammer5:
Yeah, somehow there's a difference in behavior in v33 and v32... But as I won't backport to v32 anyways, I think it should be fine to mimic the current behavior of v33 and then fix all the bugs in there

Mohammer5:
> Is the negative margin necessary on the more options button? It's tiny, but just using the right margin looks fine to me and seems less magic-numbery.

@ismay Well, I'd agree. So far I just copied the styles of the original and I'd like to keep it as close to the original as possible. I guess it'd make sense to do a proper design refactoring on this app to address all the small issues (like negative margins, etc.)

> Maybe you could lift up the definition of the react-final-form mock to a common __mocks__ folder.

I thought about this as well, but then I have to unmock all the export that I don't want to be mocked (See: https://github.com/facebook/jest/issues/936). I could move this to it's own file and make it a function call, just wondering where to put it :thinking: 

* [x] On selecting a file and clicking the file select again, I get a spinner. That's a bit jarring, and doesn't seem to happen in the version that's on play.
* [ ] I like snapshot tests for presentational components, it's useful to be notified that it changed with a diff of the changes. But for testing any component behaviour I prefer asserting that it behaves as expected. So with for example the MoreOptions component I'd prefer using enzyme to test if the elements are/aren't there. Same for click handlers, etc. Otherwise you have to manually scan the snapshot and that's easy to get wrong.
* [ ] I'd go with a unitless line-height. Seems simpler to me to have it just be proportional.
* [x] The more options button doesn't have a cursor: pointer.
* [x] Is the negative margin necessary on the more options button? It's tiny, but just using the right margin looks fine to me and seems less magic-numbery.
* [ ] Maybe you could lift up the definition of the react-final-form mock to a common __mocks__ folder.
* [x] I know we do this a lot in our apps, but I'd remove the console.logging in production.
* [x] Maybe formfooter and formheader instead of formfoot and formhead?

Mohammer5:
>  I like snapshot tests for presentational components, it's useful to be notified that it changed with a diff of the changes. But for testing any component behaviour I prefer asserting that it behaves as expected. So with for example the MoreOptions component I'd prefer using enzyme to test if the elements are/aren't there. Same for click handlers, etc. Otherwise you have to manually scan the snapshot and that's easy to get wrong.

I totally agree with this for the MoreOptions and for RadioGroup. For the components in `src/components/inputs` I wouldn't do this as it's basically testing RadioGroup again, what do you think?

ismay:
> For the components in src/components/inputs I wouldn't do this as it's basically testing RadioGroup again, what do you think?

Yeah I agree. I'd stick to snapshots there. It's good to know that it renders well, and renders as expected.

Mohammer5:
@ismay I've gone through the list of suggestions and I think I'm done with all of them. Can you have a look again?

package.json
```diff
@@ -19,22 +19,27 @@
     "@dhis2/d2-i18n-extract": "^1.0.6",
     "@dhis2/d2-i18n-generate": "^1.0.18",
     "@dhis2/ui": "^1.0.0-beta.15",
+    "@dhis2/ui-core": "^3.7.1",
     "axios": "^0.19.0",
     "classnames": "^2.2.5",
     "d2": "^30.0.4",
+    "enzyme-to-json": "^3.4.0",
```
ismay: Doesn't really matter, but probably best in devDependencies

src/App.js
```diff
@@ -63,8 +67,8 @@ class App extends React.Component {
                 loaded: true,
             })
         } catch (e) {
-            console.log('/api/me error')
-            console.log(e)
+            isProduction() && console.log('/api/me error')
```
ismay: Shouldn't this be `!isProduction()`?

Mohammer5: oh... yeah!

src/helpers/env.js
```diff
@@ -0,0 +1 @@
+export const isProduction = () => process.env.NODE_ENV === 'production'
```
ismay: Not that it matters, but was just curious why you're exporting a function instead of a boolean here.

Mohammer5: hmm good question, don't really have a reason.
I guess it's a habit of putting everything into functions to ensure immutability, but it's obviously not needed in this case, won't matter as webpack should strip it out (I assume?)

ismay: Don't know. I know that uglifyjs handles part of the dead code elimination, but wouldn't be surprised if it doesn't do that with functions.

ismay: Could grep the production build for the `isProduction console.log` lines to be sure.

Mohammer5: just checked, it get's stipped out, it's just left in the `.js.map` file

Mohammer5: Ok, it doesn't, changed it now

PR #58, opened 2019-08-07
refactor(import meta data): use final-form instead of inheritance pattern

Mohammer5:
This will also fix some bugs introduced with #56 
* I didn't include the `firstRowIsHeader` field
* I didn't parse errors properly


jason-p-pickering:
@Mohammer5 any reason you removed the Russian translations?

src/components/Inputs/ClassKey.js
```diff
@@ -0,0 +1,42 @@
...
+            <Select
+                name={CLASS_KEY_KEY}
+                label={i18n.t('Class key')}
+                options={options}
+                defaultValue={defaultValue}
+            />
+        )}
+
+        <FormSpy
+            render={({ values, form }) => {
```
ismay: They've hidden this well in the docs. `form.change` seems so much nicer than those hideous mutator functions

Mohammer5:
> @Mohammer5 any reason you removed the Russian translations?

@jason-p-pickering no! That's a mistake!

src/components/Inputs/ClassKey.js
Mohammer5: Yes!

src/pages/import/MetaData/helper.js
```diff
@@ -0,0 +1,189 @@
...
+    [MERGE_MODE_KEY]: MERGE_MODE_DEFAULT_VALUE,
+    [FLUSH_MODE_KEY]: FLUSH_MODE_DEFAULT_VALUE,
+    [SKIP_SHARING_KEY]: SKIP_SHARING_DEFAULT_VALUE,
+    [SKIP_VALIDATION_KEY]: SKIP_VALIDATION_DEFAULT_VALUE,
+    [ASYNC_KEY]: ASYNC_DEFAULT_VALUE,
+    [INCLUSION_STRATEGY_KEY]: INCLUSION_STRATEGY_DEFAULT_VALUE,
+}
+
+export const onSubmit = (setLoading, setError) => values => {
+    try {
```
ismay: Which command are we guarding against here? If possible it'd be nice to wrap it just where things might fail, so it's more clear?

ismay:
Oh one thing I noticed is the resetOnMount props that seem unused. Seems like nice functionality, but is it a placeholder atm.?

src/pages/import/MetaData/helper.js
Mohammer5: I have no idea.. Just copied the existing code. I assume it's in case of a malformed response

Mohammer5:
@ismay where did you find `resetOnMount`?

ismay:
> @ismay where did you find `resetOnMount`?

Here for example: https://github.com/dhis2/import-export-app/pull/58/files#diff-fb96b18249cb092518e19db7d8c93ae9R30

Mohammer5:
@ismay Oh yeah, you're right, I wanted to abstract the functionality that's currently in `FirstRowIsHeader`, but didn't do it (I'm more or less collecting info how to set up the `ui-forms` lib).
WIll remove it

Mohammer5:
Done

src/components/Inputs/ImportStrategy.js
```diff
@@ -0,0 +1,29 @@
+import i18n from '@dhis2/d2-i18n'
+import React from 'react'
+import { RadioGroup } from '../FinalFormComponents/RadioGroup'
+
+export const OPTION_CREATE_AND_UPDATES = {
+    value: 'CREATE_AND_UPDATES',
```
ismay: This is `CREATE_AND_UPDATE` in the version on play

src/pages/import/Data.js
```diff
@@ -59,6 +60,11 @@ export const DataImport = () => {
                             <File name="upload" />
```
ismay: So I assume this isn't a constant because it isn't reused in the app in other places?

src/components/Inputs/ImportStrategy.js
Mohammer5: Fixed

src/pages/import/Data.js
Mohammer5: Hmm good catch, I'll create an `Upload` component in `src/components/Inputs/`

ismay:
Left a couple of comments. After those this looks good for merging to me. I tested the UI, which works as expected for me (with a play/dev backend on macos chrome).

It doesn't behave exactly like the version on play, because the version on play hangs after an import. So that's fixed with this PR. But due to that bug I can't verify that there aren't any regressions in the UI.

I did test whether the POST contains the same data, and checked if the used keys are the same, and they are.

PR #60, opened 2019-08-27
fix(data import): add missing skip audit request parameter

Mohammer5:
I forgot to add the `skipAudit` value to the actual request.
Should be fixed with this PR

### DIFF
--- a/src/helpers/form.js
+++ b/src/helpers/form.js
@@ -10,6 +10,8 @@ import {
 } from '../components/Form'
 import { isValueNil } from '../helpers'
 
+const identity = value => value
+
 export function getField(name, fields) {
     return fields.filter(f => f.name === name)[0]
 }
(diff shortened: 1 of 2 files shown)


src/helpers/form.js
```diff
@@ -49,10 +49,12 @@ export function getFieldValue(field) {
 }
 
 export function getParamsFromFormState(state, list, append = []) {
-    const params = list.map(k => `${k}=${encodeURIComponent(state[k])}`)
+    const params = list.map(k =>
```
ismay: Maybe this is because one letter variables are also used elsewhere in the app, but just for readability a slightly more elaborate name is nice I think. Make it just that much easier to skim the code if the name is descriptive (unless it clutters it unnecessarily).

ismay: Same with the `v` vars below

Mohammer5: Replaced, please review this one

Mohammer5:
> To me it seems appropriate to have controls like these as switches, instead of radio buttons, for consistency (I believe that's what we do elsewhere right?).

@ismay and I have agreed that this is a separate PR

PR #61, opened 2019-08-28
chore: add debug module and a few debug logs for import request helper

Mohammer5:


ismay:
I might be commenting on something that was already decided in the notes (I'll look to make sure). If it's already been decided and we want to keep it like this then disregard my comment. My comment is also based on what @Mohammer5 and I talked about earlier today, where we were talking about debug's lack of log levels, grouping, etc. That is something they do by design, and I feel that not approaching this as a logging tool, but as a debugging tool is how this library's meant to be used. The current approach feels too much like how you'd implement a kind of logging for all apps to me.

Mohammer5:
> I see this lib as a replacement for putting console.log statements to debug something. That's something I do with a specific area of code in mind.

That's exactly what I was trying to achieve:
With having different name spaces, you can choose what specifc area of code you want to debug

I think I'd go with their recommendation of using the lib name (or app or equivalent) as the top level. 

See what @varl and I eventually agreed to in terms of top level namespaces:
https://github.com/dhis2/notes/issues/46#issuecomment-524319695

> This feels too much like it tries to anticipate a use-case.

To me it feels like it's the opposite (see reply to first quote in this comment)

Mohammer5:
> When implemented you'll be flooded with debug messages.

That was my concern as well with too much depth of namespaces / too many debug statements.
@varl eventually convinced me with:

> As you said, it depends if the developer starts top-down or bottom-up.
> (https://github.com/dhis2/notes/issues/46#issuecomment-524311652)

ismay:
Yeah I see what you mean. I remember that discussion and also gave a thumbs up in the end so I shouldn't complain.

But in general this does seem very verbose to me and I still wonder if we're not trying to anticipate the use-case too much. You might reach a point where it's still easier to put a console.log statement. To me part of the appeal is the simplicity of the lib and its usage. But I understand the points made in the other discussion.

Mohammer5:
@ismay can you explain what you would've done differently?

ismay:
@Mohammer5 That's a good question 😋. To be honest, other than the top-level namespaces I don't think I'd know where to simplify it. And since the top level namespaces were already discussed and decided on in the other issue I think that this would also be how I'd implement it.

If we're all sure on the top level namespaces with regards to what I said above, this is all right with me 👍Mostly happy to have this kind of debugging 😊

ismay:
> Of course this is still the experimental phase for using debug (I've never used debug for anything but private projects, I never got a second opinion on how I do stuff). It might be a bit cumbersome to write debug statements at this level of detail unit necessary (although you could also just do it when necessary and then keep them)

Yeah exactly. I think we'll just have to see what it's like to use it in practice. For me I also mostly used it on private projects. I think my comments were also based on me being used to how I use it there. Viktor's use-case is just as valid, so it should also accommodate that of course.

src/helpers/debug.js
```diff
@@ -0,0 +1,2 @@
+import createDebug from 'debug'
```
varl: If we are going to standardise on `debug` then this helper should likely be in a `@dhis2/log` library.

src/helpers/xhr.js
```diff
@@ -1,9 +1,14 @@
-import { getMimeType } from './mime'
-import { eventEmitter } from '../services'
+import { debug } from './debug';
 import { emitLogOnFirstResponse, fetchLog } from '../pages/import/helpers'
+import { eventEmitter } from '../services'
+import { getMimeType } from './mime'
+
+const debugHelperXhr = debug.extend('helper:xhr')
 
 // eslint-disable-next-line max-params
```
varl: This ignore should be on L9 since L10 that has a `max-params` violation.

varl: How did this get past ESLint?

src/helpers/xhr.js
```diff
@@ -21,10 +26,13 @@ export function getUploadXHR(url, upload, type, onResponse, onError, format) {
 }
 
 // eslint-disable-next-line max-params
+const debugReadyStateChange = debugHelperXhr.extend('ready-state')
```
varl: Same as above.

src/pages/import/helpers.js
```diff
@@ -8,20 +12,36 @@ const CATEGORY_2_LABEL = {
...
-        eventEmitter.emit('log', {
+        const emitData = {
             id: uid,
             d: new Date(time),
             type: `${level} - ${CATEGORY_2_LABEL[category]}`,
             text: message,
-        })
+        }
+
+        debugEmit(`Emit log, data: %O`, emitData)
```
varl: This doesn't need to be a template literal.

src/pages/import/helpers.js
```diff
@@ -32,12 +52,15 @@ export function emitLogOnFirstResponse(xhr, importType) {
...
-            eventEmitter.emit('log', {
+            const emitErrorData = {
                 id: new Date().getTime(),
                 d: new Date(),
                 type: `ERROR - ${CATEGORY_2_LABEL[importType]}`,
                 text: typeReports[0].objectReports[0].errorReports[0].message,
-            })
+            }
+
+            debugFirstResponse(`Emit log (first response), error: %O`, emitErrorData)
```
varl: Unnecessary template literal.

src/pages/import/helpers.js
```diff
@@ -50,23 +73,29 @@ export function emitLogOnFirstResponse(xhr, importType) {
...
-        eventEmitter.emit('log', {
+        const emitLogData = {
             id: new Date().getTime(),
             d: new Date(response.created),
             type: `INFO - ${CATEGORY_2_LABEL[importType]}`,
             text: message,
-        })
+        }
+
+        debugFirstResponse(`Emit log, data: %O`, emitLogData)
```
varl: More template literals that should be a string literal, seems to be a trend so won't comment individually on them.

src/pages/import/helpers.js
```diff
@@ -89,121 +118,154 @@ function isFetchLogComplete(data) {
...
         type: `INFO - ${CATEGORY_2_LABEL[type]}`,
         text: `Imported: ${imported}, Updated: ${updated}, Deleted: ${deleted}, Ignored: ${ignored}`,
-    })
+    }
+
+    debugLogCount(`Log import count: %O`, logData)
+    eventEmitter.emit('log', logData)
 }
 
+const debugLogConflicts = debugLog.extend('stats')
```
varl: Not sure about all these `extend`:ed loggers in the same file. It's a bit messy.

src/helpers/xhr.js
varl: Scratch that, looks like the lint build step caught it.

varl:
> Yeah I see what you mean. I remember that discussion and also gave a thumbs up in the end so I shouldn't complain.

💭 Changing your mind when you get new input is wiser than doubling down on your first statement just to avoid contradicting yourself.

varl:
I feel the need to point out there was no formal decision in https://github.com/dhis2/notes/issues/46 to go with `debug`. We reached an agreement about how `debug` could be implemented and what our needs from it would be.

The way this PR would have looked without that discussion would have been very different.

Now in turn, this PR informs that discussion as it shows an actual concrete implementation of how it could look in practice.

src/pages/import/helpers.js
Mohammer5: At first, I tried it without extending inside the file, which makes the file/debug code look cleaner.
But unlike on a cli, objects won't be printed as a string in the browser's console, so having all the debug call's namespaces printed in the same color makes it very hard to get a "quick" overview what ran in which order, which is why I added all the `extend`s.

I wouldn't mind reverting this, but I suggest you try both versions yourself as well.
(For cli scripts I def. wouldn't have done this)

Mohammer5:
Let's continue this discussion in https://github.com/dhis2/notes/issues/46 and take this PR as an example we can reference

src/pages/import/helpers.js
varl: Agreed. To me this is a signal that maybe `debug` isn't the right tool for the job.

varl:
Closing this to keep the PR section clean. This being closed does not in any way signify that the discussion has been settled.

PR #62, opened 2019-08-29
 fix: handle unknown api response shape

Mohammer5:
When an event file is used with 0 event, currently no summary is shown as the shape of the response is not handled by the app.

This PR will introduce handling of that response shape and show an error to the user in the summary.

### DIFF
--- a/src/components/TaskSummary/index.js
+++ b/src/components/TaskSummary/index.js
@@ -2,11 +2,19 @@ import React, { Fragment } from 'react'
 import i18n from '@dhis2/d2-i18n'
 import { eventEmitter } from '../../services'
 import { Loading } from '../'
-import { Totals, TypeStats, Conflicts, Messages, Summaries } from './helpers'
+import {
+    Errors,
+    Totals,
+    TypeStats,
+    Conflicts,
+    Messages,
+    Summaries,
+} from './helpers'
 import s from './styles.module.css'
 
 const initialState = {
     loading: false,
+    errors: '',
 
     stats: {
         created: 0,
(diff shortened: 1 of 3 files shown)


src/components/TaskSummary/index.js
```diff
@@ -2,11 +2,19 @@ import React, { Fragment } from 'react'
...
+    TypeStats,
+    Conflicts,
+    Messages,
+    Summaries,
+} from './helpers'
 import s from './styles.module.css'
 
 const initialState = {
     loading: false,
+    errors: '',
```
varl: It looks like `errors` (a string) and `error` (an array) is mixed up through out.

Mohammer5: Thanks

src/components/TaskSummary/index.js
```diff
@@ -45,7 +53,7 @@ export class TaskSummary extends React.Component {
 
     onLoaded = () => this.setState({ loading: false })
 
-    onLoading = () => this.setState({ loading: true })
+    onLoading = () => this.setState({ loading: true, error: [] })
```
varl: Are you using `error` from the state or should this be removed now that you set the `errors` array in `initialState`?

Mohammer5:
> Please take a look at the errors vs. error situation in TaskSummary/index.js.

@varl fixed

src/components/TaskSummary/index.js
Mohammer5: This should unset the errors when loading starts again, because you won't care about the errors from the previous upload, right?

varl: Sure, but then it should be `errors` (plural) and not `error` (singular).

Mohammer5: Good catch, this time there's not `error` left (except for the event, which is correct)

PR #64, opened 2019-08-30
fix: handle unknown api response shape (DHIS2-7207)

Mohammer5:
* fix(summary conflict): check length against 0 instead of array itself


Mohammer5:
Postponed until 2.32 feeze is over

PR #67, opened 2019-09-04
feat(attributes): prepare attributes fetching & state management

Mohammer5:
Preparing DHIS2-7495

* Dynamic options are now displayed on the export Data page

<hr />
Mentioning @Philip-Larsen-Donnelly so you stay up2date

### DIFF
--- a/src/helpers/api.js
+++ b/src/helpers/api.js
@@ -11,26 +11,23 @@ export const initApi = d2Instance => {
     api = d2.Api.getApi()
 }
 
-export const getAttributes = ({
-    filter = [],
-    paging = false,
-    unique = true,
-}) => {
+export const getUniqueAttributes = ({ type }) => {
     const params = [
-        `paging=${paging}`,
+        `paging=false`,
         `fields=${['id', 'displayName'].join(',')}`,
-        ...[...filter, `unique:eq:${unique}`].map(f => `filter=${f}`),
+        `filter=unique:eq:true`,
+        `filter=${type}Attribute:eq:true`,
     ]
 
     return api.get(`${ATTRIBUTES_ENDPOINT}?${params.join('&')}`)
 }
 
 export const getUniqueDataElementAttributes = () =>
-    getAttributes({ filter: ['dataElementAttribute:eq:true'] })
+    getUniqueAttributes({ type: 'dataElement' })
         .then(({ attributes }) => attributes)
         .catch(() => [])
 
 export const getUniqueOrganisationUnitAttributes = () =>
-    getAttributes({ filter: ['organisationUnitAttribute:eq:true'] })
+    getUniqueAttributes({ type: 'organisationUnit' })
         .then(({ attributes }) => attributes)
         .catch(() => [])
--- a/src/helpers/api.js
+++ b/src/helpers/api.js
@@ -1,0 +1,36 @@
+const ATTRIBUTES_ENDPOINT = 'attributes.json'
+
+let d2
+let api
+
+/**
+ * Sets d2 and the api
+ */
+export const initApi = d2Instance => {
+    d2 = d2Instance
+    api = d2.Api.getApi()
+}
+
+export const getAttributes = ({
+    filter = [],
+    paging = false,
+    unique = true,
+}) => {
+    const params = [
+        `paging=${paging}`,
+        `fields=${['id', 'displayName'].join(',')}`,
+        ...[...filter, `unique:eq:${unique}`].map(f => `filter=${f}`),
+    ]
+
+    return api.get(`${ATTRIBUTES_ENDPOINT}?${params.join('&')}`)
+}
+
+export const getUniqueDataElementAttributes = () =>
+    getAttributes({ filter: ['dataElementAttribute:eq:true'] })
+        .then(({ attributes }) => attributes)
+        .catch(() => [])
+
+export const getUniqueOrganisationUnitAttributes = () =>
+    getAttributes({ filter: ['organisationUnitAttribute:eq:true'] })
+        .then(({ attributes }) => attributes)
+        .catch(() => [])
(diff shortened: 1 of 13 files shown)


src/helpers/api.js
```diff
@@ -0,0 +1,36 @@
...
+
+export const getAttributes = ({
+    filter = [],
+    paging = false,
+    unique = true,
+}) => {
+    const params = [
+        `paging=${paging}`,
+        `fields=${['id', 'displayName'].join(',')}`,
+        ...[...filter, `unique:eq:${unique}`].map(f => `filter=${f}`),
```
ismay: I think that this is a little difficult to scan. Maybe moving a part of it out into the function body would help

src/index.js
```diff
@@ -20,36 +21,33 @@ const { REACT_APP_DHIS2_BASE_URL } = process.env
 
 init({
```
ismay: Is there a reason that we don't do this in `componentDidMount`? That way we could already render a spinner. I assume you're just sticking to the existing approach here?

src/reducers/attributes/index.js
```diff
@@ -0,0 +1,44 @@
...
+
+    dataElement: [],
+    organisationUnit: [],
+}
+
+export default function attributesReducer(
+    state = initialState,
+    { type, payload } = {}
+) {
+    if (type === LOADING_ATTRIBUTES_START) {
```
ismay: This is also nice actually, vs the standard switches in reducers.

ismay:
Looks good to me. Tested and working atm. with a 2.32.1 backend.

Mohammer5:
Thanks for reviewing @ismay.
I've added another commit so that options on the export Data page, the dynamic options will be loaded and shown. This is quite hacky and more of a quick fix, as these pages should be refactored as well..

src/index.js
Mohammer5: I've just copied our existing approaches so far.
I also want to reduce the amount of work to a minimum because this one is time critical

src/reducers/attributes/index.js
Mohammer5: yeah, the app already does it this way, but I also started using this approach.
I like to take `lodash/fp`s `isEqual function and do this in my personal projects:

```js
import { isEqual } from 'lodash/fp'

const reducer = (state, { type, payload }) => {
  const isType = isEqual(type)

  if (isType(ACTION_TYPE)) {
    return ...
  }

  return state
}
```

ismay:
I noticed a visual difference btw, the older app (on 2.32.1) looks a bit better. Old version:

Screenshot 2019-09-05 at 11 40 28

Version from this PR:

Screenshot 2019-09-05 at 11 40 35

I assume this wasn't introduced in this PR. Do you know if this is a known bug? I'll report it if not.

Mohammer5:
@ismay I think it's not been reported yet, thanks for raising the issue

src/pages/export/Data.js
```diff
@@ -36,46 +43,101 @@ export class DataExport extends FormBase {
...
+        })
+    }
 
     async componentDidMount() {
+        this.props.fetchDataElementAttributes()
+        this.props.fetchOrganisationUnitAttributes()
         await this.fetch()
     }
 
+    initializeFormValues(fieldValuesOverride) {
```
ismay: I think this method, and componentDidUpdate could use some comments on what's going on exactly. They're the areas where it's not clear to me what's happening and why.

ismay:
One other thing I now noticed, when I configure an uncompressed json export with a small number of org units and the default dates, it does not trigger a network request when I click export. I assume that that's supposed to trigger a request?

src/pages/export/Data.js
Mohammer5: I've realized I don't have to set the fields and state after the options have been loaded, so I just reverted that. I've added comments to componentDidUpdate

Mohammer5:
@ismay thanks for looking through this again, I've made some changes, please look again.
I've also reduced the "complexity" of the `src/helpers/api.js` file as there was no need to make it as dynamic as it was prior to my last change

> One other thing I now noticed, when I configure an uncompressed json export with a small number of org units and the default dates, it does not trigger a network request when I click export. I assume that that's supposed to trigger a request?

I can't reproduce this, can you give me exact details how to reproduce this?

> Oh, and shall I report this issue?

Exactly

ismay:
> I can't reproduce this, can you give me exact details how to reproduce this?

I'm using debug.dhis2.org/2.32.1 as a backend and the latest commit from this branch (fee4713). Then follow these steps:

* Go to Data Export page
* Uncollapse 'Sierra Leone' > 'Bo' > 'Badjia' and select 'Ngelehun CHC'
* For children select 'Selected organisation unit'
* Leave everything else on default, except compression to 'Uncompressed'
* Click 'Export' button
* No network request and app does nothing

ismay:
I don't think that's regression though. After testing on the import export app on debug.dhis2 it does the same there. So still a bug it seems to me, but not a regression I think.

Mohammer5:
> * Go to Data Export page
> * Uncollapse 'Sierra Leone' > 'Bo' > 'Badjia' and select 'Ngelehun CHC'
> * For children select 'Selected organisation unit'
> * Leave everything else on default, except compression to 'Uncompressed'
> * Click 'Export' button
> * No network request and app does nothing

That's because you didn't select a data set, a `required` error message would make sense here though.. (hopefully soon, with RFF)

ismay:
> That's because you didn't select a data set, a required error message would make sense here though.. (hopefully soon, with RFF)

Yeah exactly. I was bamboozled by the Select All checkbox, which is actually a button.

ismay:
Created:

https://jira.dhis2.org/browse/TECH-210
https://jira.dhis2.org/browse/TECH-211

To deal with the issues we encountered above

PR #76, opened 2019-09-10
fix(metadata dependency): use objectType and objectList in url

Birkbjo:


src/pages/export/MetaDataDependency.js
```diff
@@ -74,13 +74,20 @@ export class MetaDataDependencyExport extends FormBase {
...
+            const {
+                objectType,
+                objectList,
+                format,
+                compression,
+                sharing,
+            } = this.getFormState()
 
             const endpoint = `metadata`
+            const endpointPath = `${objectType}/${objectList}/${endpoint}`
```
Mohammer5: Should be called `endpoint` and `endpoint` could be `endpointSuffix`

PR #78, opened 2019-09-20
test: add tests for dynamic attribute options & shared id scheme attributes

Mohammer5:
* I've added cypress & cucumber/gherkin
* I've added tests for dynamic attribute options of of the data element & org unit id scheme inputs (only on the data import/export pages)
* I've added tests for id scheme input attribute options that are shared between the data element and org unit id scheme inputs

There's some "Old school" test code, which will be removed once the export pages have been refactored as well, which is why I didn't try to extract anything there as it will be thrown away anyways


cypress/integration/DataElementIdSchemeWithAttributes.feature
```diff
@@ -0,0 +1,14 @@
+Feature: Offer attributes as options in data element id scheme input
+
+    All unique attributes assiciated with data elements should
```
vilkg: associated

Mohammer5: Fixed

cypress/integration/DataElementIdSchemeWithAttributes.feature
```diff
@@ -0,0 +1,14 @@
+Feature: Offer attributes as options in data element id scheme input
+
+    All unique attributes associated with data elements should
+    be available as options in the data element id scheme input
+
+    Scenario: An attribute is associated with data elements when importing data
+        Given the user is on the data import page
+        And an attribute is associated with data elements
```
Mohammer5: This should be happening before the `Given` step above and actually stub the route

PR #79, opened 2019-09-24
refactor(meta data export): use final form & ui-core components

Mohammer5:


Mohammer5:
@ismay I think this one is ready for review.

ismay:
Ok, looking now

src/components/Inputs/Schemas.js
```diff
@@ -0,0 +1,135 @@
...
+                    />
+                </div>
+            ))}
+        </div>
+    </div>
+)
+
+const SchemasInput = ({
+    excludeSchemas,
+
```
ismay: empty line

src/components/Inputs/Schemas.js
```diff
@@ -0,0 +1,135 @@
+import { Button } from '@dhis2/ui-core'
```
ismay: This file contains a lot of components, would be more understandable if it were broken up, say along the lines of the SelectAll, SelectNone controls and the components display the actual schemas for example.

src/components/Inputs/Schemas.js
ismay: We should have a lint rule for this.

src/components/Inputs/Schemas.js
```diff
@@ -0,0 +1,135 @@
...
+
+    schemasLoading,
+    schemasLoaded,
+    schemaGroups,
+    schemaGroupLables,
+    schemaGroupOrder,
+    fetchSchemas,
+}) => {
+    useEffect(() => {
+        !schemasLoaded && fetchSchemas(excludeSchemas)
```
ismay: Shouldn't you guard against this fetching schemas when schemas are already being loaded (so thake schemasLoading into account)?

I'd personally prefer an if instead of a logical operator here by the way, I find it more readable.

ismay: Ah no of course not, strike that first comment 😋

src/components/Inputs/Schemas.js
```diff
@@ -0,0 +1,135 @@
...
+        return <Loading />
+    }
+
+    return (
+        <div className={s.container}>
+            <Controls schemaKey={SCHEMAS_KEY} />
+
+            <div className={s.formControl}>
+                {schemaGroupOrder.map(groupKey => {
+                    const label = schemaGroupLables[groupKey]
```
ismay: schemaGroupLables -> schemaGroupLabels

src/components/FinalFormComponents/Checkbox.js
```diff
@@ -0,0 +1,27 @@
+import { useField } from 'react-final-form'
+import { Checkbox as CoreCheckbox } from '@dhis2/ui-core'
+import React from 'react'
+import propTypes from 'prop-types'
+
+export const Checkbox = ({ label, name, checkedInitially: initialValue }) => {
+    const { input } = useField(name, {
```
ismay: I missed this in the previous PRs, but I was wondering why you use `useField` here. I remember that in the readme for react-final-form-hooks it was mentioned that there are certain downsides to using the hooks (see top of readme https://github.com/final-form/react-final-form-hooks). I wonder if that still goes for the hooks here? In general maybe it's nice for us to standardise on defaulting to a certain approach? To prevent divergence between our implementations?

src/components/Inputs/Schemas.module.css
```diff
@@ -2,64 +2,45 @@
...
+    user-select: none;
+    color: #000;
+    font-size: 13px;
+    font-weight: 600;
+    text-transform: uppercase;
 }
 
-.group .schema > div label {
-    margin-left: 4px;
+form .formControl {
```
ismay: I think here `column-count: 2;` would convey your intentions as well (if I assume correctly that you want to render checkboxes in two columns past a certain width). Without the risk of overflowing the container if there's too much content to fit in two 1100 pixels high columns.

src/helpers/api.js
```diff
@@ -3,6 +3,9 @@ const ATTRIBUTES_ENDPOINT = 'attributes.json'
 let d2
 let api
 
+export const getD2 = () => d2
```
ismay: This area feels weird to me. With it relying on `initApi` being run, modifying a local var and then reexporting that. What do you think? (The d2 export seems unused as far as I could see btw)

src/pages/export/MetaData.js
```diff
@@ -1,62 +1,65 @@
...
+import { Compression } from '../../components/Inputs/Compression'
+import { Sharing } from '../../components/Inputs/Sharing'
+import {
+    EXCLUDE_SCHEMAS,
+    supportedFormats,
+    initialValues,
+    onSubmit,
+} from './MetaData/helper'
+import stylesForm from '../../components/Form/styles.module.css'
+import stylesFormBase from '../../components/FormBase/styles.module.css'
```
ismay: I'd use formStyles and formBaseStyles here. Otherwise it reads as if it's forms instead of styles.

src/pages/export/MetaData/helper.js
```diff
@@ -0,0 +1,80 @@
...
+    const schemaParams = Object.keys(schemas)
+        .filter(s => schemas[s])
+        .map(name => `${name}=true`)
+        .join('&')
+    const downloadUrlParams = `skipSharing=${skipSharing}&download=true&${schemaParams}`
+    const url = `${
+        api.baseUrl
+    }${endpoint}.${endpointExtension}?${downloadUrlParams}`
+
+    window.location = url
```
ismay: If I'm not mistaken, since we're using react router you can use the history package here. That'll allow the app to route instead of the browser reloading the url and assets.

src/reducers/schemas/selectors.js
```diff
@@ -1,42 +1,140 @@
...
+ * Group and sort schemas
+ * ======================
+ */
+const groupSchemas = schemas =>
+    schemas.reduce((groups, schema) => {
+        const { group } = schema
+
+        return {
+            ...groups,
+            [group]: groups[group] ? [...groups[group], schema] : [schema],
```
ismay: This could use a comment, or be separated into multiple lines to simplify reading it.

src/reducers/schemas/selectors.js
```diff
@@ -1,42 +1,140 @@
...
+}
+
+/**
+ * Get group label helpers
+ * =======================
+ */
+function breakOnCamelCase(schemaName, name) {
+    const temp = schemaName.substr(0, name.length).replace(/([A-Z]+)/g, ' $1')
+    return temp[0].toUpperCase() + temp.substr(1)
+}
```
ismay: This could also use an explanatory comment

src/reducers/schemas/selectors.js
```diff
@@ -1,42 +1,140 @@
 import { createSelector } from 'reselect'
```
ismay: I personally think it makes sense to keep reducers and selectors close (as in same file), https://twitter.com/dan_abramov/status/730933179511640064, based on Dan's recommendations (not my own idea). In general I think what's going on with the selectors is a bit abstract, so whatever you can do to make this friendlier to newcomers to the codebase would be nice (as in explanatory comments, etc.).

src/reducers/schemas/thunks.js
```diff
@@ -0,0 +1,45 @@
+import { get, filter, map, pipe } from 'lodash/fp'
```
ismay: I think using the lodash/fp stuff here complicates this bit of code more than it simplifies it. Those who don't know this have to look it up, plus you're introducing an extra dependency and not everyone might know how to use it properly. The benefit of slightly terser code does not seem worth that to me.

src/components/Inputs/Schemas.module.css
Mohammer5: I didn't write this code.. But the way it works now, there can also be three columns.
Although I really don't like the way this is solved because all instances with a higher item amount than could fit into three cols would looks not good here.

I think we should @cooper-joe about how to do this properly..
(@cooper-joe All the available meta data schemas are listed as checkboxes here, so there could be a lot more than currently displayed at some point... But @varl said we shouldn't put time into refactoring the import-export app now. I was just working on this while waiting for progress on another import-export-app jira issue)

src/components/FinalFormComponents/Checkbox.js
Mohammer5: It's actually explained here:
https://github.com/final-form/react-final-form-hooks#whats-the-difference-between-react-final-form-hooks-and-the-hooks-introduced-in-react-final-form-v5

After reading this, I think we want to standardize using react-final-form. We'll have all the benefits when using that lib's hooks as well, so we don't need to worry about it.

Personally I prefer the hook approach over the react component approach, it's more readable to me (and it separates setting up a field and rendering a field)

src/components/Inputs/Schemas.js
Mohammer5: I don't think we should have a rule for this.
Sometimes it's really handy to group props with spaces (especially when there are like 10+ args)

src/reducers/schemas/selectors.js
Mohammer5: Yeah I had my difficulties with these functions as well... I just copied them from the existing code base and now - after reading them again - I have no clue what's actually happening here.. 
I'll try to refactor this a little and add comment so it's clear what's happening

src/components/Inputs/Schemas.module.css
ismay: > I didn't write this code.. 

Ah I see, the diff made me think this was new code.

> But the way it works now, there can also be three columns.
Although I really don't like the way this is solved because all instances with a higher item amount than could fit into three cols would looks not good here.

Yeah exactly, and for me those overflowed the container. To me it makes sense to just use 2 cols (maybe responsively add more, or less)

> But @varl said we shouldn't put time into refactoring the import-export app now. I was just working on this while waiting for progress on another import-export-app jira issue

Ah ok, makes sense 👍

src/components/Inputs/Schemas.js
ismay: Yeah I think our approaches differ in that regard. I always find the whitespace between props, imports, etc. a bit distracting, to me it hinders scanning. Sometimes I find it hard to tell what they were grouped by, as there's no explicit comment. But ok, not the most important thing in the world.

src/components/FinalFormComponents/Checkbox.js
ismay: > I think we want to standardize using react-final-form.

Yeah I think that that'd be good 👍

src/components/Inputs/Schemas.module.css
Mohammer5: Just checked, column-count doesn't work here as it's ignoring the div and divides the last group into two columns, I'll keep the current solution once we've found something better

src/pages/export/MetaData/helper.js
Mohammer5: This will simply trigger a download, it doesn't redirect to another url nor does it cause any asset to be reloaded

src/components/Inputs/Schemas.js
Mohammer5: I've split it into 1. Schema, 2. Schema/Controls and 3. Schema/SchemaGroup

src/components/Inputs/Schemas.js
Mohammer5: Done

src/helpers/api.js
Mohammer5: I've removed them completely, the onSubmit function of the MetaData export page, which was the only one that was using these, is now async

src/reducers/schemas/selectors.js
Mohammer5: done

src/reducers/schemas/selectors.js
Mohammer5: I tried to improve this, can you check if it's more understandable now? @ismay

src/reducers/schemas/selectors.js
Mohammer5: I think we should address this in a different PR

src/reducers/schemas/thunks.js
Mohammer5: Removed and rewritten

Mohammer5:
I have fixed the checkbox coloring issue as well @ismay

src/pages/export/MetaData/helper.js
ismay: Ah, I assumed this was navigating.

src/reducers/schemas/selectors.js
ismay: Yeah it seems like this wasn't introduced in this PR, so that makes sense 👍

ismay:
Nice, thanks for addressing the comments. It looks good to me! If you don't mind I'll do the functional test tomorrow morning, I'm a bit braindead atm. 🧠

ismay:
I encountered something whilst testing that I'm not sure is intentional. When selecting these options (with debug dev backend):

Screenshot 2019-10-09 at 15 13 22

What I'm getting back is metadata.xml. I'd expect to get back metadata.json.

Mohammer5:
@ismay Thanks, fixed

PR #157, opened 2019-11-14
chore: update cypress, modernize test structure & fix data test id of select

Mohammer5:


package.json
```diff
@@ -14,7 +14,12 @@
     "prebuild": "npm run localize",
     "pretest": "npm run localize",
     "lint:style": "d2-style js check --all",
-    "cypress:open": "cypress open"
+    "cypress:start": "BROWSER=none yarn start",
+    "cypress:envtest": "node ./scripts/check_env_vars.js",
+    "cypress:browser": "cypress open --config baseUrl=http://localhost:3000,defaultCommandTimeout=15000",
+    "cypress:ci": "NO_COLOR=1 cypress run -b $BROWSER --config baseUrl=http://localhost:3000,defaultCommandTimeout=15000",
+    "cypress:open": "yarn cypress:envtest && concurrently --kill-others -n cra,cypress 'yarn cypress:start' 'wait-on http-get://localhost:3000 && yarn cypress:browser --env LOGIN_URL=$REACT_APP_DHIS2_BASE_URL'",
```
varl: What does the `-n cra,cypress` switch do and what does the names map to?

scripts/check_env_vars.js
```diff
@@ -0,0 +1,131 @@
+const fs = require('fs')
```
varl: Would be nice to look into creating a lib soon so we don't need to duplicate this everywhere.

package.json
Mohammer5: it's for prefixing the output of concurrently. The first command (`'yarn cypress:start'`) will be prefixed with "cra" and the second one (`'wait-on http-get://localhost:3000 && yarn cypress:browser --env LOGIN_URL=$REACT_APP_DHIS2_BASE_URL'`) with "cypress"

scripts/check_env_vars.js
Mohammer5: Yes I absolutely agree!

package.json
Mohammer5: I chose `cra` because the import export app is a create-react-app. Maybe `yarn start` is better?

varl: I think it's fine. I didn't know if it was bound to the scripts in `package.json` or something.

PR #333, opened 2020-01-05
Add cypress tests for all import & export endpoints

Mohammer5:
This tests all the import- and export pages' import and export endpoints' parameters to make sure the pages are working as intended. I've added data-test props to all the input components.

If you want to run all tests, the `yarn cypress:open` might eat up all your RAM, so it's better to use `yarn cypress:run` (e. g. `env REACT_APP_DHIS2_BASE_URL=https://debug.dhis2.org/dev yarn cypress:run`), 

You might see that on the Data, Event & Metadata Dependency export pages, I'm also providing a "label" in the tests to select the values in the tests. I'm doing this for two reasons:

1. To know whether the component that should be selected/changed is an old/new component (with label means old, which means it'll be removed once everything is refactored; I'm preparing a PR right now with the refactored pages, so won't be long until all that crap can be removed)
2. The MUI selects don't have their values reflected on the DOM elements, so I need the value in order to select them

<hr />

@ismay It'd be nice if our new Select components could add the values to the dom elements (`data-value`?), so these can be clicked.

<hr />

@varl I think we really need a `dhis2/cypress` which abstracts the deps and other misc stuff. It also should provide support (cypress terminology) functions for interacting with our `ui-core`/`ui-widgets` components (like selecting an option by name + value (and label, if there are multiple options with the same value))


ismay:
>     2. The MUI selects don't have their values reflected on the DOM elements, so I need the value in order to select them
> 
> @ismay It'd be nice if our new Select components could add the values to the dom elements (`data-value`?), so these can be clicked.
> 
> like selecting an option by name + value (and label, if there are multiple options with the same value)

Just so I understand, because I'm not sure I follow the explanation for the labels here exactly. Why is it that the options don't always have associated labels?

ismay:
One other thing I noticed btw. is the data-test attributes. Since this is an app, and not a library, I can imagine it's not necessary to follow the convention we've used in ui-core, which namespaces with `dhis2-uicore-<componentname>`. Do we want to standardise on how we handle the data-test naming for apps? Maybe you can elaborate on the naming strategy you've settled on?

Mohammer5:
> Just so I understand, because I'm not sure I follow the explanation for the labels here exactly. Why is it that the options don't always have associated labels?

@ismay They always have labels, just when retrieving the select elements in the tests, you'd need the label only when there are multiple options with the same value, otherwise the value should be enough to select the right option.

I've written some helper functions for cypress in this PR, like `selectSelect`, which also needs a label when trying to change the value of a MUI select. So currently I need to compare the text value of the option elements to a string rather than just retrieving the one I need by using the `[name="${...}"]` selector

Mohammer5:
> One other thing I noticed btw. is the data-test attributes. Since this is an app, and not a library, I can imagine it's not necessary to follow the convention we've used in ui-core, which namespaces with dhis2-uicore-<componentname>. Do we want to standardise on how we handle the data-test naming for apps? Maybe you can elaborate on the naming strategy you've settled on?

Good point, I'll change them for now to `dhis2-import-export-*`. We could start a discussion in the `notes` repo, what do you think? @ismay

ismay:
To update others reading this issue, the data-value and data-label attributes have been added to the ui-core selects.

> Good point, I'll change them for now to `dhis2-import-export-*`. We could start a discussion in the `notes` repo, what do you think? @ismay

Sounds good, I'll start an issue in notes. By the way, the convention we settled on in ui-core omitted the dashes, so doing the same here would be dhis2-importexport-componentname, etc.

PR #336, opened 2020-01-09
Refactor export pages

Mohammer5:
With this PR all pages use RFF. Obviously the code is far from perfect and still contains many now unused modules/files, but it should be much easier to add/change/remove inputs and to refactor anything now.

This PR will also simplify test code as the tests don't have to handle two different input component types anymore (like old and ui-core radios/selects/etc)

Running the tests confirms that the endpoints are still called with the correct parameters.
Tests can be run like this:
```
env REACT_APP_DHIS2_BASE_URL=https://debug.dhis2.org/dev yarn cypress:run
```


.env.development
```diff
@@ -1,3 +1,4 @@
 # Default admin/district authorization for development
 REACT_APP_DHIS2_BASE_URL="http://localhost:8080"
 REACT_APP_DHIS2_AUTHORIZATION="Basic YWRtaW46ZGlzdHJpY3Q="
+SKIP_PREFLIGHT_CHECK=true
```
ismay: Is this line necessary? I don't have it in my local .env, and the tests run without errors for me.

src/components/FinalFormComponents/RadioGroup.js
```diff
@@ -22,39 +25,20 @@ const RadioComponent = ({ name, label, value, defaultValue }) => {
...
-
-export const RadioGroup = ({
-    label,
-    name,
-    options,
-    defaultValue,
-    dataTest,
-    ...rest
-}) => {
+export const RadioGroup = ({ label, name, options, defaultValue, dataTest, ...rest }) => {
```
ismay: `rest` is unused here right?

src/pages/export/Data.js
```diff
@@ -1,299 +1,357 @@
...
+DataExport.path = '/export/data'
+DataExport.menuIcon = <DataIcon />
+DataExport.order = 7
+DataExport.title = i18n.t('Data Export')
+DataExport.desc = i18n.t(
+    'Export data values. This is the regular export function which exports data to the DHIS 2 exchange format called DXF 2.'
+)
 
-export { ConnectedDataExport as DataExport }
+//class _DataExport extends FormBase {
```
ismay: This commented out bit can be removed right?

src/pages/export/Event.js
```diff
@@ -1,197 +1,76 @@
...
-            this.getFormState(),
-            ['programs', 'includeDeleted', 'idScheme'],
-            append
-        )
-
-        download(api.url(path) + '?' + params)
-    }
 }
+
+EventExport.dataTest = 'export-event'
```
ismay: Stylewise it might be nice to not assign to the main export, but export them separately, like in ui-core. But if you agree, I'd be fine with dealing with that in a separate PR.

src/reducers/dataSets/index.js
```diff
@@ -0,0 +1,46 @@
...
+    error: '',
+    data: [],
+    loaded: false,
+    loading: false,
+    filter: '',
+}
+
+export default function dataSets(
+    state = dataSetsDefaultState,
+    { type, payload } = {}
```
ismay: The reducer will never be called without an action right? Or can that happen?

src/components/Inputs/ObjectList.js
```diff
@@ -0,0 +1,92 @@
...
+    getObjectsLoading,
+} from '../../reducers/object/selectors'
+
+export const OBJECT_DEFAULT_OPTIONS = []
+export const OBJECT_KEY = 'objectList'
+export const OBJECT_DEFAULT_VALUE = null
+
+const objectLabel = i18n.t('Object')
+
+const RenderComponent = ({ values }) => {
```
Birkbjo: I'm still not really too familiar with hooks yet, is it normal to use components just for their side-effects like this? Seems kind of unnatural to me, especially with the name. Took me a bit to understand what's going on and why it's done this way. 
The onChange-callback on FromSpy would be more suiting I think, but then there would be more code in ObjectList.

Birkbjo: To my understanding there's a reason for this, as when I tried to replace with a simple callback, it will re-render and "forget" about the selected value. So I guess this is a way to (re)load the values on first render and when the objectType changes?

src/components/Inputs/ObjectList.js
```diff
@@ -0,0 +1,92 @@
...
+                    dataTest="input-object-list"
+                />
+            </Field>
+        )
+    }
+
+    return (
+        <>
+            <FormSpy
+                subscription={{ values: { objectType: true } }}
```
Birkbjo: Related to the comments above.
I tried to look into using a callback for this, and I ended up noticing that this seems wrong according to the API. It says `{ [string]: boolean }`, so you would think it would use a path (like elsewhere in RFF), eg. `{'values.objectType': true }`, but that does not work either. 
Currently it rerender on every value-change, but the dependency in `RenderComponent` stops it from re-fetching.
If you do the nested path, it will only call the callback/render-func on the first render.

.env.development
Mohammer5: I removed it

src/pages/export/Event.js
Mohammer5: Yeah, I'd prefer to do that in a dfferent PR

src/reducers/dataSets/index.js
Mohammer5: Removed the confusing default value

src/components/Inputs/ObjectList.js
Mohammer5: Completely changed. I've send you a draft on slack before, but please have a look again

src/components/Inputs/ObjectList.js
Mohammer5: This is now `subscription={{ values: true }}`

Mohammer5:
@ismay @Birkbjo I think I've addressed all your concerns/questions. Please have a look again

PR #386, opened 2020-01-28
fix: assign download url correctly

Mohammer5:


Mohammer5:
> The Data section doesn't seem to do anything at all, I click the export-button but nothing happens, no console or network errors... just nothing...

This is something that's really stupid and I haven't addressed yet (although I did refactor this part, I didn't touch the "existing" bugs).
Both the org unit and the data set inputs are required. Only then you can submit the form. But this is not new, so I won't address this here

> The Event section has an enabled submit button but when I click that, I get to see an ugly error page*. After I select an org-unit, all is well

Same here... I agree that this is terrible.. But I won't put too much effort into this app anymore because there's a complete refactor right now, and until there's a decision about what's going to happen, this might be "wasted" effort (unless @varl agrees that this has some kind of priority because this app wil be around until something happens and this app really s**ks the way it is..)

Mohammer5:
Actually it should be fairly simple to add this... Will take a look after fixing this issue

PR #418, opened 2020-03-23
chore: conditional rendering return statement simplified

awgaan:
Following @Mohammer5 suggestion on return statement logic and putting `App.js` in line with the rest of the code base.


awgaan:
Agreed. Since we're rendering `CssReset` unconditionally now, the fragment keeps the code DRY.

PR #423, opened 2020-03-24
feat: optimize api schema request

awgaan:
only get the fields we need

### DIFF
--- a/cypress/integration/EndpointParamChecks/ExportMetaData/index.js
+++ b/cypress/integration/EndpointParamChecks/ExportMetaData/index.js
@@ -1,7 +1,7 @@
 import '../common/settingFormValues'
 import { Before, Given, Then } from 'cypress-cucumber-preprocessor/steps'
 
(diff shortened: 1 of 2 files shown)


cypress/integration/EndpointParamChecks/ExportMetaData/index.js
```diff
@@ -1,7 +1,7 @@
 import '../common/settingFormValues'
 import { Before, Given, Then } from 'cypress-cucumber-preprocessor/steps'
 
-const schemasApi = /api\/schemas.json\?fields=metadata,collectionName,displayName,klass/
+const schemasApi = /api\/\d{2}\/schemas\?fields=metadata,collectionName,displayName,klass/
```
Mohammer5: @varl is there a chance that we'll ever have a different version in the api path other than two digits? (Probably in 35 years, but I'm certain no line of code we currently have will survive that long)

varl: This is in test code, so I'm not too worried about it.

PR #424, opened 2020-03-24
fix(ui): button clarity

awgaan:
Select/clear all buttons are now `Button`s instead of `Checkbox`es.

## New
image

## Old
image


cooper-joe:
>What do you think @cooper-joe?
Yeah, much better controls for selecting/deselecting all 👍

ismay:
Mayb instead of setting TZ, you could mock the Date object.

PR #546, opened 2020-05-05
feat(event import): add data-element id scheme & id scheme input fields

Mohammer5:
This PR

* changes the event id scheme field type from radio buttons to a dropdown field.
* adds the `Data Element Id Scheme` & `Id Scheme` input fields to the event import page

image


ismay:
Couple of linting errors:

Screenshot 2020-05-05 at 15 40 36

amcgee:
Ref https://jira.dhis2.org/browse/DHIS2-8714

amcgee:
@Mohammer5 looks like this fixes it for event import but not for export?  https://jira.dhis2.org/browse/DHIS2-8714 refers to both

PR #557, opened 2020-05-11
feat: add schemes for event import export

awgaan:
Event import:
- Add id scheme selection input

Event export:
- Add data element scheme selection input
- Add org unit scheme selection input

relevant Jira issue: DHIS2-8714

### DIFF
--- a/cypress/integration/EndpointParamChecks/ExportEvent.feature
+++ b/cypress/integration/EndpointParamChecks/ExportEvent.feature
@@ -5,16 +5,18 @@ Feature: The user should be able to export events
         Given the user is on the event export page
         And the more options are visible
         And the following options are set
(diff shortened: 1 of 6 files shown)


awgaan:
Ordering it based on relation makes sense. Changed.
I'll fix the cypress issue in another branch.

Mohammer5:
>  I'll fix the cypress issue in another branch.

Could we do that before we approve this PR? I'd like to run the tests as an extra verification step until we can automate that

awgaan:
> > I'll fix the cypress issue in another branch.
> 
> Could we do that before we approve this PR? I'd like to run the tests as an extra verification step until we can automate that

 #566 should fix the Cypress issue.

PR #559, opened 2020-05-12
chore(deploy): fix deploy-build to upload the correct directory

amcgee:
deploy-build uploads the `build` directory by default, but for App Platform apps it needs to upload `build/app` - see https://github.com/dhis2/data-visualizer-app/blob/master/.travis.yml#L14

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ script:
     - npm run build
 deploy:
     - provider: script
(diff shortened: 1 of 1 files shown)


amcgee:
As an aside, every commit to every branch should be "published" to d2-ci, so you can see the effect of this change at https://github.com/d2-ci/import-export-app/tree/chore/deploy-build-dir

PR #560, opened 2020-05-12
feat: scheme inputs use data query instead of manual fetch

awgaan:
Custom fetch functions replaced with `useDataQuery`.
Reason: `app-runtime` now supports multiple filters.

### DIFF
--- a/cypress/integration/DynamicAttributes/IdSchemeWithCommonAttributes/index.js
+++ b/cypress/integration/DynamicAttributes/IdSchemeWithCommonAttributes/index.js
@@ -1,7 +1,5 @@
 import { Given, Then } from 'cypress-cucumber-preprocessor/steps'
 
-const loginUrl = Cypress.env('dhis2_base_url')
-
 const getKeyBySide = side => (side === 'org units' ? 'orgUnit' : 'dataElement')
 
 Given('the user is on the data {word} page', type => {
@@ -17,24 +15,24 @@ Given(
 
         if (left === 'org units') {
             cy.route({
-                url: `${loginUrl}/api/attributes.json?paging=false&fields=id,displayName&filter=unique:eq:true&filter=organisationUnitAttribute:eq:true`,
+                url: /api\/\d{2}\/attributes\?filter=unique:eq:true&filter=organisationUnitAttribute:eq:true&fields=id,displayName&paging=false/,
                 response: 'fx:orgUnitAttributes',
             }).as('attributesXHR')
 
             cy.route({
-                url: `${loginUrl}/api/attributes.json?paging=false&fields=id,displayName&filter=unique:eq:true&filter=dataElementAttribute:eq:true`,
+                url: /api\/\d{2}\/attributes\?filter=unique:eq:true&filter=dataElementAttribute:eq:true&fields=id,displayName&paging=false/,
                 response: {
                     attributes: [],
                 },
             })
         } else {
             cy.route({
-                url: `${loginUrl}/api/attributes.json?paging=false&fields=id,displayName&filter=unique:eq:true&filter=dataElementAttribute:eq:true`,
+                url: /api\/\d{2}\/attributes\?filter=unique:eq:true&filter=dataElementAttribute:eq:true&fields=id,displayName&paging=false/,
                 response: 'fx:dataElementAttributes',
             }).as('attributesXHR')
 
             cy.route({
-                url: `${loginUrl}/api/attributes.json?paging=false&fields=id,displayName&filter=unique:eq:true&filter=organisationUnitAttribute:eq:true`,
+                url: /api\/\d{2}\/attributes\?filter=unique:eq:true&filter=organisationUnitAttribute:eq:true&fields=id,displayName&paging=false/,
                 response: {
                     attributes: [],
                 },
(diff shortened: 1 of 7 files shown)


src/components/ElementSchemes/DataElementIdScheme.js
```diff
@@ -1,32 +1,40 @@
...
+import { useDataQuery } from '@dhis2/app-runtime'
 import PropTypes from 'prop-types'
 import i18n from '@dhis2/d2-i18n'
 
-import { fetchAttributes } from '../../utils/helper'
 import { dataElementIdSchemeOptions } from '../../utils/options'
 import { SelectField } from '../'
 
+const schemeQuery = {
+    data: {
```
Mohammer5, line 10: I think `data` is not a good identifier, it could mean anything. I'd prefer to just call it `dataElements` (Don't see a need for abbreviating identifiers here)

src/components/ElementSchemes/DataElementIdScheme.js
```diff
@@ -1,32 +1,40 @@
...
+    const { loading, error, data } = useDataQuery(schemeQuery)
 
     const validationText =
         error &&
         `${i18n.t(
             'Something went wrong when loading the additional data element ID schemes'
         )} : ${error.message}`
 
+    const schemes = data
+        ? data.data.attributes.map(({ id, displayName }) => ({
```
Mohammer5, line 30: `data.data` looks weird, see above

src/components/ElementSchemes/IdScheme.js
```diff
@@ -1,63 +1,80 @@
...
+    },
+    data: {
+        resource: 'attributes',
+        params: {
+            filter: ['unique:eq:true', 'dataElementAttribute:eq:true'],
+            fields: ['id', 'displayName'],
+            paging: 'false',
+        },
+    },
+}
```
Mohammer5, line 26: Same as above

src/components/ElementSchemes/IdScheme.js
```diff
@@ -1,63 +1,80 @@
...
+    const sharedAttributes = dataElementAttributes.reduce(
+        (shared, attribute) => {
+            const foundInOrgUnits = attributeFoundIn(
+                attribute,
+                organisationUnitAttributes
+            )
+            return foundInOrgUnits ? [...shared, attribute] : shared
+        },
+        []
+    )
```
Mohammer5, line 58: Not a biggie, but wouldn't it make more sense to first find the shared attributes and then map to the actual option format? Then it doesn't need to be done individually for both the data element and the org unit attributes

src/components/ElementSchemes/OrgUnitIdScheme.js
```diff
@@ -1,32 +1,40 @@
...
+import { useDataQuery } from '@dhis2/app-runtime'
 import PropTypes from 'prop-types'
 import i18n from '@dhis2/d2-i18n'
 
-import { fetchAttributes } from '../../utils/helper'
 import { orgUnitIdSchemeOptions } from '../../utils/options'
 import { SelectField } from '../'
 
+const schemeQuery = {
+    data: {
```
Mohammer5, line 10: Same as above

src/components/ElementSchemes/IdScheme.js
awgaan: Yeah, it makes more sense. This code is copied from the pre-refactor version. Will change.

awgaan:
Have now renamed identifiers and changed the `findSharedAttributes` function.

cypress/integration/DynamicAttributes/IdSchemeWithCommonAttributes/index.js
```diff
@@ -17,24 +15,24 @@ Given(
 
         if (left === 'org units') {
             cy.route({
-                url: `${loginUrl}/api/attributes.json?paging=false&fields=id,displayName&filter=unique:eq:true&filter=organisationUnitAttribute:eq:true`,
+                url: /api\/\d{2}\/attributes\?filter=unique:eq:true&filter=organisationUnitAttribute:eq:true&fields=id,displayName&paging=false/,
```
Mohammer5, line 18: This (and the following changes) break tests against the `debug` backend as the root url contains a path `/dev` (contrary to `/` when using localhost). This will def. be the case when doing proper e2e tests in the future, so this should def. be considered, I guess just adding back the `loginUrl` should suffice.

Mohammer5: This (and all the other requests) are happening like this:

```
filter=unique:eq:true,organisationUnitAttribute:eq:true
```

I don't know which version is correct, but the test doesn't match the request.

awgaan: I think the `app-shell` causes this issue.
It still depends on an older version of `app-runtime`.

Removing the `.d2` and `node_modules` folder and reinstalling the dependencies makes all the tests pass. I think we can wait for the `app-shell`s dependencies to be updated before merging this branch.

amcgee: @awgaan you shouldn't need to upgrade the dependency in `app-shell`, just upgrade it in the app's dependency tree - `yarn upgrade @dhis2/app-runtime`

amcgee: If there's a local `app-runtime` dep it means you have two copies which also won't work, but you can run `npx yarn-deduplicate --packages @dhis2/app-runtime` as well if needed

amcgee: There's only one version of `app-runtime` in your `yarn.lock` and it's the latest, so the `app-runtime` version isn't the issue here https://github.com/dhis2/import-export-app/blob/081146381f99f6a81939b8fd981dab516a227db2/yarn.lock#L1317

PR #661, opened 2020-09-04
fix: TEI export format

awgaan:
The trackedEntityInstances endpoint does not respect the `format`
parameter and requires us to append the format to the resource name
when making the request.

Addresses:
- DHIS2-9404 (Missing format extension in the TEI export url)
- DHIS2-9403 (Malformed URL when exporting teis with programStatus=ALL)


amcgee:
Is this a backend regression?  If so we should raise it with the backend rather than hacking it here!

awgaan:
Not a regression.

awgaan:
Is closed by  #665.
Lots of non UX commits there to put it mildly.

PR #663, opened 2020-09-04
fix(i18n): correctly import translations

awgaan:
Forgot to import from `./locales`.

Addresses:
- DHIS2-9402 (Translations are not supported in import-export app)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import { HashRouter } from 'react-router-dom'
-import i18n from '@dhis2/d2-i18n'
+import i18n from './locales'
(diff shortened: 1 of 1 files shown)


amcgee (approved):
LGTM

amcgee:
The deploy here fails because `i18n/ur-translatsions` is a branch name that already exists in `d2-ci` and `git` doesn't support pushing both a ref and a ref folder to the same location (`i18n`) - see [this notes issue](https://github.com/dhis2/notes/issues/100).  I will delete the `ur-translations` branch from `d2-ci` (since it's 2 years old) but in general it would be good to use longer, more specific and descriptive branch names cc @awgaan

PR #674, opened 2020-09-10
feat(ui): prefix labels with keywords

awgaan:
Fixes the label of the `DELETE` (import) strategy and shows a warning when `DELETE` is selected. Also adds back the technical terms as a prefix for most radio group inputs.

Alerts are now positioned correctly fixed in the center of the screen.

The organisation unit tree now has a max height of `512px`.

Addresses:
- DHIS2-9470 (Label describing DELETE strategy is incorrect)
- DHIS2-9472 (Alert is positioned incorrectly)


Birkbjo (approved):
LGTM

cooper-joe:
Including the tech terms for org.unit selection creates an issue, we are trying to avoid using the term "CHILDREN" and "DESCENDANTS" when referring to org.units (it is too loaded with meaning from Child programs in Tracker). We could replace these:

`Children` -> `Directly Below`
`Descendants` -> `All Below`

cooper-joe:
Also the ALLCAPS tech terms are difficult to read/scan, allcaps text is objectively harder to parse. Having so many of these labels makes an issue. I'll report it as a separate issue though if I come up with a solution.

awgaan:
@cooper-joe should the label also then be?:
```
DIRECTLY BELOW: Include the first level of units inside selections
```

cooper-joe:
yeah,
`DIRECTLY BELOW: Include the first level of units inside selections` and `ALL BELOW: Include all units inside selections`.

Note that there is a ticket to adjust the wording of these labels, but I will follow up that on jira separately.

PR #676, opened 2020-09-10
fix: sync tracked entity instance import

awgaan:
Fixes synchronous tracked entity instances import.
Also limit the max height of type report messages; apparently this can contain thousands of messages.

### Explanation
Synchronous tracked entity instances imports return a report inside the `response` property and not the `typeReports` property.


awgaan:
The code for import data/summary extraction should be refactored.
See #677 . But that requires a significant time-consuming effort.

Birkbjo:
LGTM. We can indeed look into refactoring it later, but don't worry to much about it - it does seem like the API has a few edge-cases and special behaviours so it is expected that the consuming code is complex.

Birkbjo (approved):

PR #678, opened 2020-09-11
feat(export): disable export button while content is loading

awgaan:
When compression is selected we disable the export button for 5 seconds
reason: no way of knowing how long it actually takes the server to respond

Otherwise disable export button until server has responded.

Addresses:
- DHIS2-9488 (Export: missing user feedback when processing)


Birkbjo (approved):
LGTM

PR #686, opened 2020-09-16
fix(export): show circular loader until download prompt appears

awgaan:
Disabling the export button for an arbitrary 5s regardless of how long it takes the server to respond doesn't really address the initial issue.

`locationAssign` now creates an iframe and leaves the export button disabled until the iframe has loaded which in our case also triggers a browser download prompt to appear.

The whole page also shows a circular loader while we're waiting for the server to respond.

Addresses:
- DHIS2-9488 (Export: missing user feedback when processing)

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2020-09-11T17:40:48.035Z\n"
-"PO-Revision-Date: 2020-09-11T17:40:48.035Z\n"
+"POT-Creation-Date: 2020-09-17T12:32:26.012Z\n"
+"PO-Revision-Date: 2020-09-17T12:32:26.012Z\n"
 
 msgid "Something went wrong when loading the current user!"
 msgstr ""
@@ -830,6 +830,9 @@ msgid "Job started."
 msgstr ""
 
 msgid "An unknown error occurred. Please try again later"
+msgstr ""
+
+msgid "Loading exported data"
 msgstr ""
 
 msgid "Data"
--- a/src/pages/DataExport/DataExport.js
+++ b/src/pages/DataExport/DataExport.js
@@ -72,6 +72,7 @@ const DataExport = () => {
             title={PAGE_NAME}
             desc={PAGE_DESCRIPTION}
             icon={PAGE_ICON}
+            loading={!exportEnabled}
             dataTest="page-export-data"
         >
             <Form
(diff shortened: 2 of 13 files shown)


cooper-joe (changes requested):
I'm testing locally against `https://verify.dhis2.org/2.35.0_rc/` and the loading spinner never disappears. Screencap:
Kapture 2020-09-16 at 15 04 40

The fullscreen loading spinner should disappear as soon as the download has begun.

awgaan:
Can confirm that this issue exists in Chrome. Hm.

awgaan:
There seems to be no way of making this solution work in Chrome.
The `onLoad` event is not triggered when the iframe's `src` is a file download.

@cooper-joe Would it be acceptable to open a new tab that closes when the server responds?
Is the browser's loading mouse icon and tab title enough?

cooper-joe:
> @cooper-joe Would it be acceptable to open a new tab that closes when the server responds?
> Is the browser's loading mouse icon and tab title enough?

This is not ideal, it would be a slightly confusing experience I think with auto-opening and closing tabs, but perhaps if it is the only technical solution. What would happen with the original page? It is important that the original page shows the loading state, it is this that prevents multiple clicks on export and general confusion.

I would need to try the solution you describe to be sure, but it sounds as if it would still be quite confusing.

awgaan:
@cooper-joe The new solution opens the export url in a new tab. The tab's title is `Loading exported data`. In Firefox there is also a centered paragraph with the same text as the title. The app shows a circular loader until the server has responded and the user is either met with a download prompt or the browser automatically starts downloading to the default download folder.

I tested this new solution with both Firefox and Chrome.

cooper-joe:
Thanks @awgaan. This works for me in chrome. The opening/closing of tabs is not the best user experience, but it is an improvement over no feedback.

An issue:
- Selecting 'uncompressed' does not open a new tab nor start a download.

How could we fix this so we don't have to open a new tab etc., could you open a backend ticket?

awgaan:
Agreed. I'll ponder on how this could be fixed in the future without requiring opening a new tab.
Some of the endpoints are being fixed by Ameen and will work after he has pushed his changes.

cooper-joe:
Ok thanks. I will let a dev take over and do a code-review for approval.

Birkbjo:
Yeah it's not ideal, but there's no proper way of tracking progress when downloading a file. The only way I can see this working "properly" is to make the api async, so we're given a link/location header with job-status, much like async-import.

Birkbjo:
LGTM. Even though I would prefer the first solution, maybe along with a message that the file is being generated and will start automatically once that is complete. Though, anecdotally it is pretty common for downloads to be opened in a new tab like this. And if @cooper-joe prefers this solution I will go along with it.

Birkbjo (approved):

cooper-joe:
> LGTM. Even though I would prefer the first solution, maybe along with a message that the file is being generated and will start automatically once that is complete. Though, anecdotally it is pretty common for downloads to be opened in a new tab like this. And if @cooper-joe prefers this solution I will go along with it.

@Birkbjo I also prefer the first solution you describe, but it did not work in chrome - see comments above. If it is possible then it is preferable to the current 'open-in-new-tab' behaviour.

PR #687, opened 2020-09-17
feat: show form validation summary when form is submitted

awgaan:
Addresses:
- DHIS2-9546 (https://jira.dhis2.org/browse/DHIS2-9546)


Birkbjo:
This seems very complicated for a pretty basic feature. You can get the `errors`-state from the `FormState` pretty easily: https://final-form.org/docs/final-form/types/FormState#errors . So I would just pass that object to the validationForm, or use `useFormState` from the `ValidationSummary`-component.

Birkbjo (changes requested):
Unless I'm missing something, this seems very complicated for a pretty basic feature. You can get the errors-state from the FormState pretty easily: https://final-form.org/docs/final-form/types/FormState#errors . So I would just pass that object to the validationForm, or use `useFormState` from the ValidationSummary-component.

awgaan:
You're absolutely right, Birk.
`useFormState` is so much simpler and achieves the same goal.
I over-complicated this issue.

Birkbjo (approved):

PR #688, opened 2020-09-18
fix(data/tei-export): send attachment parameter

awgaan:
The `dataValueSets` and `trackedEntityInstances` endpoints added support for the `attachment` parameter.
Client-side blob creation is now completely gone.

Addresses:
- DHIS2-9445 (Exported TEIs open in the browser, file is not downloaded)

See:
- DHIS2-9496 (Add download flag to dataValueSets and trackedEntityInstances)
  - https://github.com/dhis2/dhis2-core/pull/6184

### DIFF
--- a/src/pages/DataExport/form-helper.js
+++ b/src/pages/DataExport/form-helper.js
@@ -1,34 +1,28 @@
 import { locationAssign, compressionToName } from '../../utils/helper'
-import { getMimeType } from '../../utils/mime'
-import i18n from '@dhis2/d2-i18n'
 
-import { FORM_ERROR } from '../../utils/final-form'
-import {
-    createBlob,
-    downloadBlob,
-    jsDateToISO8601,
-    pathToId,
-} from '../../utils/helper'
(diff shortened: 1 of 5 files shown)


src/pages/DataExport/form-helper.js
```diff
@@ -39,52 +33,25 @@ const valuesToParams = ({
         `orgUnit=${selectedOrgUnits.map(o => pathToId(o))}`,
         `dataSet=${selectedDataSets}`,
         `format=${format}`,
-        `compression=${compressionToName(compression)}`,
-    ].join('&')
+        compression ? `compression=${compressionToName(compression)}` : '',
+        `filename=${filename}`,
```
Birkbjo, line 37: Shouldn't this be `attachment=`?

Birkbjo (approved):
Nice to get rid of the blobs! Just a minor comment/question

src/pages/DataExport/form-helper.js
awgaan: Yes. Important catch. :+1:

PR #690, opened 2020-09-21
fix(tei-export/org-unit-mode): correctly show ouMode overrides

awgaan:
Accessible, capture and ALL overrides the manual selection and the UI now reflects these overrides.

Addresses:
- DHIS2-9545 (Unclear if 'accessible' and 'capture' units overrride org.unit selection)

### Preview
image


vilkg:
Could we have the manual selection as a first item? Before capture and accessible?

awgaan:
@vilkg Manual selection is the default selected option. Do you feel like it's still important to have that as the first option?

Birkbjo (approved):

PR #696, opened 2020-09-24
fix: alert bar performance

awgaan:
AlertBar used to cause CPU spikes up to a 100% of one core.
This change fixed it though I am not sure why.

### DIFF
--- a/src/components/FormAlerts/FormAlerts.js
+++ b/src/components/FormAlerts/FormAlerts.js
@@ -1,13 +1,11 @@
-import React, { useEffect, useState } from 'react'
+import React from 'react'
 import PropTypes from 'prop-types'
 import { AlertStack, AlertBar } from '@dhis2/ui'
 
-const FormAlerts = ({ alerts, dataTest }) => {
-    const [bars, setBars] = useState([])
-
-    useEffect(() => {
-        if (alerts) {
-            const newBars = alerts.map(a => (
+const FormAlerts = ({ alerts, dataTest }) => (
+    <AlertStack data-test={dataTest}>
+        {alerts &&
+            alerts.map(a => (
                 <AlertBar
                     key={`alert-${a.id}`}
                     warning={a.warning}
@@ -18,15 +16,9 @@ const FormAlerts = ({ alerts, dataTest }) => {
                 >
                     {a.message}
                 </AlertBar>
-            ))
-            setBars(newBars)
-        }
-    }, [alerts])
-
-    if (bars.length == 0) return null
-
-    return <AlertStack data-test={dataTest}>{bars}</AlertStack>
-}
+            ))}
+    </AlertStack>
+)
 
 FormAlerts.propTypes = {
     dataTest: PropTypes.string.isRequired,


Mohammer5 (approved):
Just out of curiousity: Would it work as well to add an `else` block to the previous version and reset the bars to `[]` as you're doing in the new code? Maybe that's the issue (a never clear stack of bars)?

awgaan:
@Mohammer5 Yes, that also fixes it. I thought this way was cleaner.

HendrikThePendric (approved):
This makes a lot of sense 👍

ismay:
> This makes a lot of sense 👍

Does it? I mean, can anyone explain what the bug was to me? I'm not sure I get it.

ismay:
Maybe we should file a bug report for further investigation later on. Just in case there's something we're doing wrong that we're all missing here.

awgaan:
@ismay I don't understand the bug either.
This new implementation has the exact same performance issues as before on my machine.
But let's ignore that for now.

ismay:
> @ismay I don't understand the bug either.
> This new implementation has the exact same performance issues as before on my machine.
> But let's ignore that for now.

It does? Ok. Would you mind filing a jira issue for that? Just so we can track it, for later.

awgaan:
Yeah, doing that now.
Would never have gone for a `useEffect` `useState` solution if such a simple solution fixed this issue.

awgaan:
https://jira.dhis2.org/browse/DHIS2-9613

HendrikThePendric:
> Does it? I mean, can anyone explain what the bug was to me? I'm not sure I get it.

I meant it made a lot of sense for these two reasons:
- There was no reason for having state and `useEffect` in this component
- Having the state update in the useEffect hook would trigger an extra render every time the alerts array was updated, so removing that would theoretically improve performance by 50%.

Also, I was under the impression the performance issue did go away after simplifying this component. And to me that would be reason to just leave it there... However, I've just read this comment:
> This new implementation has the exact same performance issues as before on my machine.

So now I think that this PR shouldn't have been merged since the problem persists. But since it already has been merged there's no turning back, and filing a new issue seems like the way forward.

Also, I don't get this:
> Would never have gone for a useEffect useState solution if such a simple solution fixed this issue.

How were you expecting that adding more re-render triggers would have solved performance issues?

My expectation personally is that what you see happening in the alert component is merely a symptom of something else going wrong higher up in the component hierarchy. The AlertBar will reinitialise itself when its prop change, and this triggers a state change, reset a timer, and trigger a CSS keyframe animation to start. The same thing happens when the component mounts..... So if something is updating props very frequently on this component, I can see this causing issues.

ismay:
> I meant it made a lot of sense for these two reasons:
>
>    There was no reason for having state and useEffect in this component
>    Having the state update in the useEffect hook would trigger an extra render every time the alerts array was updated, so removing that would theoretically improve performance by 50%.

Ah I see what you mean. Yes I agree 👍

> My expectation personally is that what you see happening in the alert component is merely a symptom of something else going wrong higher up in the component hierarchy. The AlertBar will reinitialise itself when its prop change, and this triggers a state change, reset a timer, and trigger a CSS keyframe animation to start. The same thing happens when the component mounts..... So if something is updating props very frequently on this component, I can see this causing issues.

Yeah expect something like that as well. Weird that it's so hard to reproduce.

PR #704, opened 2020-09-25
fix: tei export assigned user mode

awgaan:
Adds a new default option to `Assigned user mode` that disables filtering by `assignedUserMode`.

### Reason
`assignedUserMode` set to `ANY` is not the same excluding the parameter completely from the parameters.

### DIFF
--- a/src/pages/DataExport/form-helper.js
+++ b/src/pages/DataExport/form-helper.js
@@ -52,6 +52,9 @@ const onExport = (baseUrl, setExportEnabled) => async values => {
     const downloadUrlParams = valuesToParams(values, filename)
     const url = `${apiBaseUrl}${endpoint}?${downloadUrlParams}`
     locationAssign(url, setExportEnabled)
+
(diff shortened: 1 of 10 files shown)


src/pages/DataExport/form-helper.js
```diff
@@ -52,6 +52,9 @@ const onExport = (baseUrl, setExportEnabled) => async values => {
     const downloadUrlParams = valuesToParams(values, filename)
     const url = `${apiBaseUrl}${endpoint}?${downloadUrlParams}`
     locationAssign(url, setExportEnabled)
+
+    // log for debugging purposes
+    console.log('data-export:', { url, params: downloadUrlParams })
```
Birkbjo, line 57: Should probably have a proper logger for this, so that it can be filtered etc. Fine for now though.

Birkbjo (approved):

Birkbjo (approved):

PR #707, opened 2020-09-25
fix(assigned-user-mode): shorten text and remove one header

awgaan:
### Before
image

### Now
image

PR #716, opened 2020-09-29
chore(port): update v35

awgaan:
- Updates `@dhis2/ui` to include a fix for the `AlertBar`
- Includes commits in #711 
- Includes commits in #707


awgaan:
Birk can you wait a bit with the approval?
Let's get #727, #722 and #717 in here as well.

PR #722, opened 2020-09-30
feat: add home page

awgaan:
Adds a home page (`overview`) that lists all the available export and import types with short descriptions.

### Preview
image

### Discussion
Joe, Gintare and Phil wanted the home page back.
Joe agreed on the design.


Birkbjo (approved):
LGTM. Unsure about the need for the capitalisation of the page title, but 👍

PR #982, opened 2021-02-18
fix: remove preheat option & fix cypress tests

Mohammer5:
Fixes [DHIS2-10385](https://jira.dhis2.org/browse/DHIS2-10385)

* Fixed the cypress tests (which are now all passing again)
* Removed the preheat option from the meta data import form


mediremi (approved):
LGTM :heavy_check_mark:

PR #1288, opened 2022-01-27
fix: add ADX option for data export

tomzemp:
This PR adds ADX option for data export (see https://jira.dhis2.org/browse/DHIS2-4978)

PR #1290, opened 2022-01-28
fix: add ADX option for data export

benguaraldi:
This PR adds ADX option for data export (see https://jira.dhis2.org/browse/DHIS2-4978)

It is the same as https://github.com/dhis2/import-export-app/pull/1288 by @tomzemp but since that is in an external repo, it seems we can neither run the tests nor merge it.


benguaraldi:
@mediremi, do you mind reviewing this PR?  It seems we can't merge #1288, as it was from an external repo.

mediremi (approved):

PR #1433, opened 2022-06-07
feat: new GeoJSON importer (DHIS2-7171)

turban:
This PR will add support for a new GeoJSON importer that can be used to update org unit geometries and "additional geometries" (catchment areas). This importer can replace the current (and complex!) GML importer, but we plan to keep both for the upcoming DHIS2 version. 

Fixes: https://jira.dhis2.org/browse/DHIS2-7171

After this PR a new "GeoJSON import" will be available in the left menu: 

Screenshot 2022-06-13 at 14 02 48

The user has the following options to link GeoJSON features to org units, and to import data as "associated geometries": 

Screenshot 2022-06-13 at 14 03 38

We reuse the import summary that is used for other importers: 

Screenshot 2022-06-13 at 14 04 11

Note: There is a backend bug with synchronous import at the moment. The error message returned is: 

```
{
  "uid": "Pt7NALTVaJu",
  "level": "ERROR",
  "category": "GEOJSON_IMPORT",
  "time": "2022-06-13T13:51:19.524",
  "message": "GeoJSON import failed: failed to lazily initialize a collection of role: org.hisp.dhis.user.User.userRoles, could not initialize proxy - no Session",
  "completed": true,
  "id": "Pt7NALTVaJu"
}
```

When the backend bug is fixed we will switch to asynchronous mode for this importer. 

I have tried to reuse as much code as possible, but had to make some changes: 
 
1. A new input to "Match GeoJSON property to organisation unit scheme" (`GeojsonPropertyMatch`). This input combines a checkbox with a text field and a dropdown: 

Screenshot 2022-06-13 at 14 12 29

Screenshot 2022-06-13 at 14 12 48

2. The current `OrgUnitIdScheme` component include more schemes than supported by the GeoJSON importer, so I created a new `OrgUnitIdSchemeCore` component that only supports `Id`, `Code` and `Name`: 

Screenshot 2022-06-13 at 14 19 10

3. A new input to "Import as associated geometry" (`GeojsonAttributePicker`) which reuses the `ResourcePickerField` component. This components lists all attributes  of type "GeoJSON" applied to "Organisation unit" (defined in the Maintenance app) that can be used for the import. 

Screenshot 2022-06-13 at 14 25 01

4. The `SingleSummary` component was adjusted to show GeoJSON feature indexes where there are errors (as the id property is not required):

Screenshot 2022-06-13 at 14 27 59


HendrikThePendric:
> @austin Not sure about "Organization Unit Import" as we're not importing org units (they need to exist before we import the geometry). Maybe we could name it `Org unit geometry import` or just `Geometry import` and list `GeoJSON` and `GML` as an indented sub-menu?

My preference would be `Org unit geometry import` or perhaps `catchment area import` and the submenus you described.

src/components/Inputs/GeojsonPropertyMatch.js
```diff
@@ -0,0 +1,52 @@
...
+
+const MATCH_PROPERTY_NAME = 'matchProperty'
+const MATCH_PROPERTY_LABEL = i18n.t(
+    'Match GeoJSON property to organisation unit scheme'
+)
+const MATCH_PROPERTY_DATATEST = 'input-match-property'
+
+const NAME = 'geojsonProperty'
+const DATATEST = 'input-geojson-property'
+const LABEL = i18n.t('GeoJSON property name')
```
jenniferarnesen, line 17: These potentially wouldn't get translated at runtime. Do they need to be constants or can all these just be added directly in the jsx?

src/components/Inputs/GeojsonAttributePicker.js
```diff
@@ -0,0 +1,93 @@
...
+const SELECTEDLABEL = i18n.t('Selected programs')
+const ERRORMESSAGE = i18n.t('Something went wrong when loading the attributes!')
+const RESOURCETYPE = resourceTypes.GEOJSON_ATTRIBUTE
+
+const VALIDATOR = (selectedAttributes = []) =>
+    selectedAttributes.length === 0
+        ? i18n.t('One attribute must be selected')
+        : undefined
+
+const GeojsonAttributePicker = ({
```
jenniferarnesen, line 28: @HendrikThePendric Are all input components supposed to be located here, even if the component would most likely never be used by another importer?

src/pages/GeojsonImport/GeojsonImport.js
```diff
@@ -0,0 +1,89 @@
...
+})
+
+const GeojsonImport = () => {
+    const {
+        tasks: { geojson: geojsonTasks },
+        addTask,
+    } = useContext(TaskContext)
+
+    // recreating a previously run job
+    const query = useLocation().query
```
Birkbjo, line 41:
```suggestion
    const { query }  = useLocation()
```

?

turban: Fixed.

src/components/Inputs/GeojsonAttributePicker.js
HendrikThePendric: Yes

src/components/Inputs/GeojsonAttributePicker.js
```diff
@@ -0,0 +1,93 @@
...
+    multiSelect,
+    label,
+    withFilter,
+    ...rest
+}) => {
+    const { input: checkbox } = useField(USE_ATTRIBUTE_NAME)
+    const { value: useAttribute } = checkbox
+
+    return (
+        <div style={{ maxWidth: '420px', paddingBottom: '24px' }}>
```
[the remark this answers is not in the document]
HendrikThePendric: `SchemeContainer` seems to be intended specifically to wrap a number of select components and restrict their width. It has a different maxWidth and doesn't have a `paddingBottom`. So I think it's fine to just use a div here.

HendrikThePendric (changes requested):
This looks very good. Thanks for adding this and aligning with the code-style in this repo. I would like to request one general change, which is that you don't apply inline styles, but use CSS modules instead.

src/components/Inputs/GeojsonAttributePicker.js
turban: I have moved the specific Geometry components to the components/Geometry folder to clean up a bit. Tell me if I should move them to the components/Inputs folder.

jenniferarnesen (approved):

HendrikThePendric (approved):

PR #1503, opened 2022-09-20
fix: dataElementGroup no longer needed to fetch current data values [v39]

jenniferarnesen:
With [this backend change](https://github.com/dhis2/dhis2-core/pull/11671), it is now possible to fetch data using the `dataValueSets` endpoint without having to provide a dataElementGroup id.

See, look, the current values are successfully fetched with this change!
image

Also in this PR: add placeholder for the Data element selector


janhenrikoverland (approved):
LGTM

PR #1507, opened 2022-09-21
fix: display ou parent name in preview with grey color

jenniferarnesen:


cooper-joe (approved):

i18n/en.pot
```diff
@@ -1101,8 +1101,14 @@ msgstr "Women 80 years and above"
 msgid "No organisation units have been selected"
 msgstr "No organisation units have been selected"
 
-msgid "No associated geometry found for selected organisation units"
-msgstr "No associated geometry found for selected organisation units"
+msgid ""
+"It is not possible to get data from Earth Engine for point facilities. "
+"Select polygon organisation units or use a catchment area as associated "
+"geometry.”"
```
edoardo, line 1107: I don't see the opening quotes, are they supposed to be somewhere or should this closing quotes be removed?

edoardo (approved):

i18n/en.pot
jenniferarnesen: It snuck in there. Removed it now.

PR #1508, opened 2022-09-22
fix: page number needs to be reset to 1 when new preview generated [v39]

jenniferarnesen:
Do not persist the page number because it may become invalid when the preview is regenerated and there are fewer rows of data. Therefore, put the pageNo into the state for the components that are removed every time the preview is generated.

Side note: the two Population Preview components could probably be refactored since about 80% of the code is identical. But that will have to be done for a later version.


martinkrulltott (approved):
Any reason for not just calling things `page` instead of `pageNo`? (just like the `Pagination` component does)

jenniferarnesen:
> Any reason for not just calling things `page` instead of `pageNo`? (just like the `Pagination` component does)

I can do that.

PR #1509, opened 2022-09-22
fix: adjust spacing so Job summary header and tags look better [v39]

jenniferarnesen:
Before: (with file uploaded, eg geoJson file import):

image

Before: (no file, as is the case with ee import)
image

After (with file uploaded, eg geoJson file import):
image

After (no file, as is the case with ee import)
image


janhenrikoverland (approved):
LG

PR #1516, opened 2022-10-03
fix: patch webpack to support webworker importScripts with relative publicPath

amcgee:
Fixes https://dhis2.atlassian.net/browse/DHIS2-13850

This uses [patch-package](https://www.npmjs.com/package/patch-package) to hot-fix webpack, adding support for "undoing" asset path to find the true root directory.  It's not an ideal solution, especially because some webpack tests are failing (perhaps they just need to be updated) but it does appear to solve the issue.

Another solution would be to patch the `webpack.config.js` config in `react-scripts` to use `publicPath: 'auto'`, but this requires further investigation to ensure it doesn't break i.e. the custom webpack plugins bundled with createReactApp.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
         "format": "yarn format:js && yarn format:text",
         "cy:start": "BROWSER=none yarn start",
         "cy:open": "d2-utils-cypress open --appStart 'yarn cy:start'",
-        "cy:run": "d2-utils-cypress run --appStart 'yarn cy:start'"
+        "cy:run": "d2-utils-cypress run --appStart 'yarn cy:start'",
+        "postinstall": "patch-package"
     },
     "devDependencies": {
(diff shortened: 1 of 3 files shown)


amcgee:
This patches webpack to fix a bug there causing issues with web workers.  It’s not ideal as a generic webpack fix (which is why I haven’t yet opened a PR against webpack - might want to do something updating the `publicPath` in Create React App to use `auto`, solving it there instead of in webpack, though that requires more work to check all the edge cases.  

However this should be very low-risk as a patch in the import/export app itself, because it only affects web workers and we only use web workers in this one place which has been tested thoroughly and works with the patch in place.

amcgee:
See also https://github.com/facebook/create-react-app/issues/12503, https://github.com/facebook/create-react-app/pull/12755, https://github.com/webpack/webpack/issues/16119

jenniferarnesen (approved):

PR #1551, opened 2022-10-19
fix: don't show list when empty [DHIS2-13971] (master)

HendrikThePendric:
Fixes [DHIS2-13971](https://dhis2.atlassian.net/browse/DHIS2-13971)


KaiVandivier (approved):
Tested & looks good :)

PR #1651, opened 2023-01-28
fix: use correct metadata update strategy [DHIS-12479]  

rithvikreddykoripelli:
Hi Team, 

Raising this PR to fix the following bug 
https://dhis2.atlassian.net/browse/DHIS2-12479

The error is due to null pointer exception in the backend when strategy is "UPDATES". As per the backend logic, it is expecting "UPDATE" as strategy.  

Thanks,
Rithvik

### DIFF
--- a/src/components/Inputs/ImportStrategy.js
+++ b/src/components/Inputs/ImportStrategy.js
@@ -16,7 +16,7 @@ const importStrategyOptions = [
         prefix: i18n.t('Append'),
     },
     {
-        value: 'UPDATES',
+        value: 'UPDATE',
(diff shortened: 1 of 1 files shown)


rithvikreddykoripelli:
Hi @tomzemp , 
I have changed the commit message and force pushed the code. I will keep this in mind next time. Thanks

tomzemp:
thanks @rithvikreddykoripelli. I'm just looking into getting our remaining CI jobs to run before approving (I'm not sure if we're able to run them on a fork with our current setup, so I'm following up on that)

PR #1951, opened 2023-11-03
feat(DHIS2-16133): migrate to new tracker API

kabaros:
This PR updates Tracker Entity and Event import/export to use the new Tracker API as described in the docs [here](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/tracker-deprecated.html).

implements [DHIS2-16133](https://dhis2.atlassian.net/browse/DHIS2-16133)

## Changes in this PR
The commit messages should give an overview of the changes, but here are some more details:

- call the new endpoints: `/api/tracker/events` and `api/tracker/trackedEntities` instead of the old `api/trackedEntityInstances` and `api/event`
- use `orgUnit` instead of `ou` for defining the organisation unit
- updates the fetching summary logic to show the report returned by the new scheduler job (used internally by tracker)
image
- remove XML for the available formats
- change `atomicMode` field to `ALL` and `OBJECT` (instead of `ALL` and `NONE` in the old API)
- Remove `none` from the options for Tracked Entity export filter, as the new API requires choosing either `program` or `tracked entity`
- Tracked Entity import can be done asynchronously now, so we're using the async option same as with other imports.
- Update UI references from "Tracked Entity Instance" to "Tracked entity"
- For events, change change `startDate` and `endDate` to `occuuredBefore` and `occuredAfter`
- For tracked entities, change `lastUpdateStartDate` to `updatedAfter`, ` lastUpdateEndDate` to `updatedBefore`, and `lastUpdatedDuration` to `updatedWithin`
- For tracked entities, change `programStartDate` to `enrollmentEnrolledAfter` (and also endDate)
- Change `followupStatus` to `followup` and ensure that the parameter is not sent if ALL is chosen (the new API expects true or false, or for it not to be provided in the case of ALL)

[DHIS2-16133]: https://dhis2.atlassian.net/browse/DHIS2-16133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tomzemp (commented):
Can we remove the XML (DXF2) references from the captions,
e.g.
image

image

tomzemp (approved):

PR #1963, opened 2023-12-12
fix: avoid breaking selects [DHIS2-16264]

KaiVandivier:
https://dhis2.atlassian.net/browse/DHIS2-16264

The select components do a fetch on render for attributes to populate some options (the top two or three are hard-coded though). If you select an option that was populated by the fetch, that value is saved in the Final Form state. If you collapse the Advanced Options section, then re-expand the section, the select tries to choose the option based on that value, _but_ the component needs to fetch again, so that option isn't available on first render

**Edited:**
I changed the `<MoreOptions>` component to keep children rendered to preserve state, and then hide them with CSS

Before:

https://github.com/dhis2/import-export-app/assets/49666798/6af20755-1c1d-4069-98f7-7ada89b0b83b

After:

https://github.com/dhis2/import-export-app/assets/49666798/503004b9-85a9-417b-8041-3ef4bac0f98f

### DIFF
--- a/src/components/ElementSchemes/DataElementIdScheme.js
+++ b/src/components/ElementSchemes/DataElementIdScheme.js
@@ -22,7 +22,7 @@ const DataElementIdScheme = ({
         fetchAttributes(`${baseUrl}/api/`, 'dataElementAttribute')
             .then((attributes) => setSchemes(attributes))
             .catch((error) => setError(error))
-        setLoading(false)
+            .finally(() => setLoading(false))
     }, [])
 
     const validationText =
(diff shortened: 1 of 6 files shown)


Mohammer5:
> There are other ways to redesign things that manage data and time-to-first-render a bit more optimally, but I think this week is about quick and simple

In that case we should add another Jira issue so we don't forget about it

Mohammer5 (approved):
LGTM, as I said, I think we should add another Jira issue so we know that the UX of the advanced options section can be improved

PR #1977, opened 2024-01-29
fix: update captions to reflect the menus below [DHIS2-12655]

benguaraldi:
This fixes the captions in the main menu and elsewhere in the Import/Export app so that they reflect the file formats that are found within.

Discussion of this between @cooper-joe and I can be found at:
https://dhis2.atlassian.net/browse/DHIS2-12655

I'm a little unsure of the best practice of changing text that is also found in the translation files, so I went ahead and altered them too.  The English text will not match the other languages where it has been specified until this text is re-translated.  Is there some notification that automatically happens with that?


Philip-Larsen-Donnelly:
@benguaraldi Unfortunately you can't change the .po files directly; they are pulled from the translation platform (transifex). In addition, I believe the source (en.pot) file is generated from the code. So, basically, the .pot and .po file changes should not be included; only the javascript need be updated.

benguaraldi:
@Philip-Larsen-Donnelly Thanks!  That's a bummer that the changing of the text invalidates the close-but-not-exact translations.

This PR now has no changes to `po` files and the autogenerated `pot` file.

tomzemp:
Thanks for the PR @benguaraldi 🙏

Looks good! Did you want to use the word `format` in the Data Import caption instead of `files` to be consistent with the others?
image

I would defer to @karolinelien or @cooper-joe on this one as I don't have strong feelings about what the language should be. As I now look at this, I do find it a bit confusing that our captions mention, for example, `DXF 2 XML` and `ADX XML` but then the options on the page will have `XML` and `ADX`. I guess we haven't been consistent before, however, and changing the options is probably going to be more confusing that changing the captions.

benguaraldi:
Good catch, @tomzemp!  I've updated that text to match the others.  Thanks!

cooper-joe:
Thanks for the PR @benguaraldi. The changes certainly make captions clearer and more accurate.

> I would defer to @karolinelien or @cooper-joe on this one as I don't have strong feelings about what the language should be. As I now look at this, I do find it a bit confusing that our captions mention, for example, DXF 2 XML and ADX XML but then the options on the page will have XML and ADX. I guess we haven't been consistent before, however, and changing the options is probably going to be more confusing that changing the captions.

Yes, it seems the captions are now more accurate, but that accuracy now highlights the unclear labelling of the options. For example:

**Data Export**
Caption: Export metadata, such as data elements and organisation units, to JSON, CSV, DXF 2 XML, or ADX XML format.
Option labels: JSON, CSV, XML, ADX

So, in the above example it seems `DXF 2 XML` is being shortened to just `XML` in the options, and `ADX XML` is being shortened to just `ADX` in the options. 

@benguaraldi, am I understanding this correctly? In this case, do you think the option names should be changed too? (to `DXF 2 XML` and `ADX XML` respectively).

benguaraldi:
@cooper-joe Thanks!  Yes, that's right—I'm using `ADX XML` to mean the same thing as `ADX` in the options and `DXF 2 XML` to refer to the same thing as `XML` in the options.

Maybe the best thing is to change both `DXF 2` and `XML` to `DXF2` as well as `ADX XML` to `ADX`?  That would remove `XML` from the captions and the options, but since we support two different types of XML, maybe that's appropriate?

(`DXF2` seems clearer to me than `DXF 2` and is how @larshelge styles it in [this COP post](https://community.dhis2.org/t/dxf-detailed-metadata-export/1427/7), for instance.)

@bobjolliffe and @jimgrace definitely know more about these XML formats than I do, so might have ideas about how to best indicate them.

benguaraldi:
@cooper-joe Okay, I've matched the options and the captions and standardized on DXF2 and ADX, omitting the XML.

I did verify that the TEI import and export are also in the DXF2 format.

Perhaps this is now good to go?

benguaraldi:
@cooper-joe That's a great solution!  Thanks!

I've implemented the XML parenthetical in the options only, not the captions, which is I think what you were looking for, but let me know if that's wrong.

cooper-joe (approved):
Thanks for the changes @benguaraldi, looks good to me. 

@dhis2/platform-frontend, need a quick dev review on this one as it now touches the frontend (label changes only, though).

tomzemp (approved):
lgtm. @benguaraldi: do you want us to merge this for you?

benguaraldi:
@tomzemp Please!  I'm not allowed to merge, apparently.  :)

PR #1978, opened 2024-02-23
chore: move app to continuous release

kabaros:
moving import/export app to continous release from v41. I didn't apply to earlier versions since it seems that a lot has changed between versions (geo import, and other stuff) so it's hard to tell what applies to all versions or not.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "import-export-app",
-    "version": "1.5.69",
+    "version": "99.9.9",
     "description": "",
     "license": "BSD-3-Clause",
     "private": true,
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "import-export-app",
-    "version": "1.5.69",
+    "version": "99.9.9",
```
Mohammer5, line 3: I guess this is so that the first release will be 100.0.0?

Mohammer5 (approved):
I don't know too much about these configuration files, changes look good to me though.
Just a single question, approving though. If you want to get a more qualified review, you could ask someone else who knows more about the release processes.

package.json
kabaros: yeah exactly - for context, here is a writeup of the steps needed to move to continuous release: https://github.com/dhis2/notes/discussions/293

PR #1985, opened 2024-02-28
feat: force trigger a release

kabaros:
BREAKING CHANGE: bump to v100 for continuous release


Chisomchima (approved):
I think it's all Good

PR #2002, opened 2024-03-28
feat: update the summary table to display a message

Chisomchima:
Implements [DHIS2-17164](https://dhis2.atlassian.net/browse/DHIS2-17164)
---

### Description
In the case of a dry run, this commit updates the table summary (screenshots below), and instead show a Notification that says "{ignoredCount} entities will be ignored out of {totalCount} entities"

---

### Known issues

-   [ ] None

---

### Checklist

-   [ ] API docs are generated
-   [ ] Tests were added
-   [ ] Storybook demos were added

_All points above should be relevant for feature PRs. For bugfixes, some points might not be relevant. In that case, just check them anyway to signal the work is done._

---

### Screenshots
Screenshot 2024-03-28 at 11 53 27

Screenshot 2024-03-28 at 15 28 56

[DHIS2-17164]: https://dhis2.atlassian.net/browse/DHIS2-17164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/JobSummary/Summary/Summary.js
+++ b/src/components/JobSummary/Summary/Summary.js
@@ -1,3 +1,5 @@
+import i18n from '@dhis2/d2-i18n'
+import { NoticeBox } from '@dhis2/ui'
 import PropTypes from 'prop-types'
 import React from 'react'
 import { typeReportParse } from '../helper.js'
@@ -6,7 +8,7 @@ import { TypeReportSummary } from '../TypeReportSummary/TypeReportSummary.js'
 import styles from './Summary.module.css'
 
 const extractStats = (summary) => {
-    if (summary.responseType == 'ImportSummaries') {
+    if (summary.responseType === 'ImportSummaries') {
         const { imported, deleted, ignored, updated, total } = summary
         return { imported, deleted, ignored, updated, total }
     } else if (summary.importCount) {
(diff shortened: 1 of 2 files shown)


src/components/JobSummary/Summary/Summary.js
```diff
@@ -18,8 +20,25 @@ const extractStats = (summary) => {
     }
 }
 
-const Summary = ({ summary, importType }) => {
-    // gml import type object return
+const Summary = ({ summary, importType, isDryRun }) => {
+    if (isDryRun) {
```
kabaros, line 24: this logic looks like it would apply to all import dry runs, but we want to only apply it to `tracked entities` and `events` .. right?

kabaros (commented):
one comment + please fix failing tests

src/components/JobSummary/Summary/Summary.js
```diff
@@ -18,8 +20,25 @@ const extractStats = (summary) => {
     }
 }
 
-const Summary = ({ summary, importType }) => {
-    // gml import type object return
+const Summary = ({ summary, importType, isDryRun }) => {
+    if (isDryRun && importType === 'TRACKER_IMPORT_JOB') {
```
kabaros, line 24: I would just put this in a variable to make the intention clearer
```suggestion
   const isTrackerDryRun = isDryRun && importType === 'TRACKER_IMPORT_JOB
    if (isTrackerDryRun) {
```

and maybe add a comment for posterity:

```
// the new tracker endpoints don't return a full summary, so we're just showing the ignored/total
```

PR #2033, opened 2024-05-20
fix: use dry run instead of import mode when making api call to old tracker api

flaminic:
Partial fix for [DHIS2-14864](https://dhis2.atlassian.net/browse/DHIS2-14864)
Description

Partial fix for dry run mode on import of TEI, changed the parameter in the requests as per [the documentation](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-240/tracker-deprecated.html) . The API Backend needs to be fixed as well for the bug to be resolved

[DHIS2-14864]: https://dhis2.atlassian.net/browse/DHIS2-14864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tomzemp (approved):
I approve, but the ticket says the backend change is only going to be backported down to 2.39? In which case, this change here won't contribute to fixing the problem in the app, so I guess it's not worth merging?

PR #2037, opened 2024-05-30
fix(DHIS2-16988): download uncompressed json rather than open inline

kabaros:
This allows downloading exported `json` files - right now, it just opens in a new window since the API no longer returns the `content-disposition` header for JSON uncompressed. To fix that, we're using the [download](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#download) property of an anchor to trigger a download in the browser. Since the download is triggered programatically, rather than through a link element, we have to create the button in JS and trigger it, similar to this [solution](https://stackoverflow.com/a/49917066). 

Testing this solution needs to happen on the same origin, so it won't work on netlify. You need to build the app and install in an instance, then it works.  
fixes [DHIS2-16988](https://dhis2.atlassian.net/browse/DHIS2-16988)

[DHIS2-16988]: https://dhis2.atlassian.net/browse/DHIS2-16988?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ps: I had a solution [here](https://github.com/dhis2/import-export-app/pull/2015) that creates a longer file name containing other info related to the export, but canned it for now as we didn't agree whether we should do this or not.

### DIFF
--- a/src/utils/helper.js
+++ b/src/utils/helper.js
@@ -188,22 +188,32 @@ const uploadFile = ({
     })
 }
 
-const downloadWindowTitle = i18n.t('Loading exported data')
-const downloadWindowHtml = `
-<div style="height: 100%; width: 100%; display: flex; justify-content: center; align-items: center; color: rgb(33, 41, 52)">
-    <p>${downloadWindowTitle}</p>
-</div>
-`
-
 // call stub function if available
-const locationAssign = (url) => {
+const locationAssign = (relativeUrl) => {
     if (window.locationAssign) {
-        window.locationAssign(url)
+        window.locationAssign(relativeUrl)
     } else {
-        const downloadWindow = window.open(url, '_blank')
+        try {
+            const url = relativeUrl.startsWith('..')
+                ? new URL(relativeUrl, document.baseURI).href
+                : relativeUrl
 
-        downloadWindow.document.title = downloadWindowTitle
-        downloadWindow.document.body.innerHTML = downloadWindowHtml // does not work in Chrome
+            const urlFilePart = new URL(url).pathname.split('/').pop()
+            const [, file, extension] = urlFilePart.match(/(^[^.]+)(\..+$)/)
+
+            const downloadedFileName = `${file}${extension}`
+
+            const link = document.createElement('a')
+            link.href = url
+            link.download = downloadedFileName
+            link.target = '_blank'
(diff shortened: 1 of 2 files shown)


src/utils/helper.js
```diff
@@ -188,22 +188,32 @@ const uploadFile = ({
...
-const downloadWindowTitle = i18n.t('Loading exported data')
-const downloadWindowHtml = `
-<div style="height: 100%; width: 100%; display: flex; justify-content: center; align-items: center; color: rgb(33, 41, 52)">
-    <p>${downloadWindowTitle}</p>
-</div>
-`
-
 // call stub function if available
-const locationAssign = (url) => {
+const locationAssign = (relativeUrl) => {
```
Birkbjo, line 192: Oh, I thought baseUrl from `useConfig` would be the absolute URL? But it may depend if the it's available from `system/info` I guess?

Very minor: but not sure if the rename of this variable is better when it _can_ be an absolute url as well?

src/utils/helper.js
```diff
@@ -188,22 +188,32 @@ const uploadFile = ({
...
+            const url = relativeUrl.startsWith('..')
+                ? new URL(relativeUrl, document.baseURI).href
+                : relativeUrl
+
+            const urlFilePart = new URL(url).pathname.split('/').pop()
+            const [, file, extension] = urlFilePart.match(/(^[^.]+)(\..+$)/)
 
-        downloadWindow.document.title = downloadWindowTitle
-        downloadWindow.document.body.innerHTML = downloadWindowHtml // does not work in Chrome
+            const downloadedFileName = `${file}${extension}`
```
Birkbjo, line 204: Is this needed? I thought the browser would handle the file-extension and we could just send the name, if we wanted?

Birkbjo (approved):
LGTM, just some questions really.

src/utils/helper.js
kabaros: as in, I can only provide the `file` part and the browser will figure out the extension? I don't think so but I will give it a try

kabaros: you're right @Birkbjo - there was no need to add the extension, removed it now and addressed your comments

PR #2038, opened 2024-05-30
fix: update captions to reflect the menus below (v40) [DHIS2-12655]

benguaraldi:
Attempting to backport #1977, which fixes the captions in the main menu and elsewhere in the Import/Export app so that they reflect the file formats that are found within.

Discussion of this between @cooper-joe and I can be found at:
https://dhis2.atlassian.net/browse/DHIS2-12655


linadhis2:
@benguaraldi Should this be closed/removed? As the ticket did get merge to both v40 and master and this one states v40, but it's into master... I verified successfully, so it seems like this may not be necessary.

benguaraldi:
@linadhis2 Yes, this PR is very wrong and is replaced by https://github.com/dhis2/import-export-app/pull/2039, which has been merged, as you noted.

I'm not sure what the best practice is—should I close this and then delete the branch?  Or something else?

tomzemp:
@benguaraldi : we can just close this as it was replaced with your other PR

benguaraldi:
@tomzemp Thanks!  And thanks for your other help too!

PR #2058, opened 2024-07-01
fix: include event programs for export [DHIS2-17672]

tomzemp:
See https://dhis2.atlassian.net/jira/software/c/projects/DHIS2/issues/DHIS2-17672

This PR fixes an issue whereby the programs query had been amended to filter out WITHOUT_REGISTRATION (event) programs, which is relevant for the Tracked Entity Export page, but not for the Event export page.

Due to the structure of the existing code, I opted to just add an additional query (without filters) and look up the appropriate within <ProgramPicker> based on a new `includeEvents` parameter.

### Automated tests
I've added a test for <ProgramPicker> to check that useDataQuery is called with the right query based on the `includeEvents` parameter. There is little test coverage in this app, currently; I think the test for this change could be written better, but it requires a bunch more setup, like setting up a query provider. 

### Manual tests

**Event export page:**
image
_note: `Antenatal care visit` which is an event program is shown_

**Tracked entity export page:**
image

_note: `Antenatal care visit` which is an event program is NOT shown_

### DIFF
--- a/src/components/Inputs/ProgramPicker.js
+++ b/src/components/Inputs/ProgramPicker.js
@@ -12,7 +12,8 @@ const LISTNAME = 'selectedPrograms'
 const FILTERLABEL = i18n.t('Filter programs')
 const SELECTEDLABEL = i18n.t('Selected programs')
 const ERRORMESSAGE = i18n.t('Something went wrong when loading the programs!')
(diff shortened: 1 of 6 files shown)


kabaros (approved):
👌🏿

Chisomchima (approved):

PR #2176, opened 2025-02-04
fix: cleanup requests to adhere to new tracker API

flaminic:
List of changes: 

- the `ouMode` parameter is used in our code in for event export, tracked entity export. this does not do anything with the new endpoints and should be replaced by `orgUnitMode`
- for events exports, the `attachment={metadata}.json` parameter does not do anything and should be removed
- it is not correct in new tracker to use fomat=json and we should instead append .json / .csv at the end
- links=false is invalid and should be removed
- On Events import / Tracked entities import / Tracked entities export,  under Advanced Options > Event ID Scheme selectors should be removed. Requests should be updated to remove references to these parameters
- On Events import / Tracked entities import / Events export / Tracked entities export, under Advanced Options > ID Scheme selectors should include the option "Name". Requests should correspondingly be updated to use Name as appropriate
- We have links to documentation for error codes, but the documentation is wrong, can quickly become out of date. Remove the link.


tomzemp (approved):
Looks good to me! 🚀

I noticed a remaining reference to ouMode in one of the test URLs. Not sure if that's intentional? https://github.com/dhis2/import-export-app/blob/d1d50a8966ea4fc9277e00e80e88e3ddbc64b8f3/src/utils/helper.test.js#L6

PR #2200, opened 2026-02-26
fix: add flexibility to coc code matching for ee import

BRaimbault:
Follow-up to [DHIS2-20774](https://dhis2.atlassian.net/browse/DHIS2-20774)

### Description

In "Earth Engine Import", when importing "Population age groups WorldPop Global2", the matching of age/sex groups relies on the "Category option combinations" (CoCs) codes, which, currently, have to be defined in the format: `F_00` / `F_10` / `M_0` / ... . However, in the case these exact codes are already in use elsewhere, it would be impossible for the admin to assign them again to the CoCs and thus would prevent the use of the matching feature (leaving them to do a teadious manual match).

This PR adds flexibility to the matching logic so imports continue to work even when codes include prefixes or suffixes, e.g., `WP_F_00` or `F_00_WP`.

Below is an extract of the metadata package we are developping to facilitate the import process where we intend to use prefixes: 
image

Note:
> It is no uncommon for instances to have multiple overlapping "Category options", "Categories", "Category combinations" and "Category option combinations" for dimensions like Age and Sex/Gender. These can be grouped via "Category option groups" for analysis. Codes are often used with prefixes or suffixes to ensure uniqueness.

### Testing

* Bundle: [import-export-999.99.9-eeimportcoc-rc3.2026-03-02.zip](https://github.com/user-attachments/files/25688630/import-export-999.99.9-eeimportcoc-rc3.2026-03-02.zip)
* Deployed at: https://dev.im.dhis2.org/wp-g2-webinar/apps/import-export#/import/earthengine
* In the app, select "Population age groups WorldPop Global2" and in data element look up "Population age/sex groups (WorldPop Global2) and check the matching is done properly.
 
image

[DHIS2-20774]: https://dhis2.atlassian.net/browse/DHIS2-20774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BRaimbault (commented):
Ready for review.

jenniferarnesen (approved):

PR #2250, opened 2026-07-16
fix: run GML import async and track it under the METADATA_IMPORT job [DHIS2-21758]

karolinelien:
## Jira
[DHIS2-21758](https://dhis2.atlassian.net/browse/DHIS2-21758) — *Import/export app UI crash / hang when GML is imported*

## Symptom
Starting an org-unit-geometry **GML** import shows "Job started." and then hangs — no progress, no summary — even when the backend import actually succeeds.

## Root cause (two issues on the GML path)
1. **The request was never made async.** `gml-helper.js` passed `isAsync: true` to `uploadFile` but left `async` **out of the request URL** (`params = \`dryRun=${dryRun}\``). So the backend ran the import **synchronously** and returned the report inline with no job `id`. `uploadFile`, told it was async, then stored the task with `id: undefined, completed: false` and waited for a job that never existed — hence the perpetual "Job started" hang. (The metadata and geojson helpers correctly include `async=${isAsync}` in the URL.)
2. **Wrong notifier category.** Once made async, `POST /api/metadata/gml` runs as a **`METADATA_IMPORT`** job (there is a declared-but-unused `GML_IMPORT` job type). The app polled `system/tasks/GML_IMPORT/<id>` / `taskSummaries/GML_IMPORT/<id>`, which return nothing, so progress/summary would still never appear.

## Fix
- `gml-helper.js`: add `async=${isAsync}` to the request params so a real async job is created and its `id` is returned.
- `useTasks.js`: map the GML task's poll category `GML_IMPORT → METADATA_IMPORT` (backend job type) for the event/summary queries only. `GML_IMPORT` stays as the UI category, so job grouping, the filter chips and "Recreate job" (`/import/gml`) are unchanged.

## Verified against play/dev (2.44) and dev-2-43 (2.43)
With an org unit whose code matches the GML feature:
- New async call `POST /api/metadata/gml.json?dryRun=false&async=true` → `200`, returns a job `id` with `jobType: METADATA_IMPORT`.
- Polling `system/tasks/METADATA_IMPORT/<id>` → streams progress → `"Import complete with status OK … 1 updated"`, `completed: true`.
- UI: import now completes with a summary instead of hanging (verified in a local build against play/dev).

Note: a `409`/`E5001` ("No matching object for reference") on this endpoint is expected and correct when no org unit matches the GML feature's uid/code/name — it is not part of this bug.

AI Assisted

[DHIS2-21758]: https://dhis2.atlassian.net/browse/DHIS2-21758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/hooks/useTasks.js
+++ b/src/hooks/useTasks.js
@@ -1,4 +1,4 @@
-import { useDataEngine } from '@dhis2/app-runtime'
+import { useConfig, useDataEngine } from '@dhis2/app-runtime'
 import { useState } from 'react'
 import { allCategories, toNotifierCategory } from '../utils/tasks.jsx'
 
@@ -46,7 +46,8 @@ const defaultJobOverview = {
 const defaultRefetchPeriod = 2000
 
 const createFetchEvents =
-    (engine, setTasks, fetchSummary) => (type, id, task) => {
+    ({ engine, setTasks, fetchSummary, apiVersion }) =>
+    (type, id, task) => {
         const fetchEvents = async () => {
             if (task.completed) {
                 return
@@ -60,7 +61,7 @@ const createFetchEvents =
 
             const response = await engine.query(query, {
                 variables: {
-                    type: toNotifierCategory(task.importType),
+                    type: toNotifierCategory(task.importType, apiVersion),
                     taskId: task.id,
                 },
             })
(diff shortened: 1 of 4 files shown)


tomzemp (approved):
@karolinelien @flaminic : in principle, I think this is good. I'm kind of tempted to just map `GML_IMPORT` to `METADATA_IMPORT` in gml-helper.js, but that does cause us to lose the ability to identify this as a "GML" job in "jobs overview", so I guess this mapping approach is better.

The only thing I'd probably ask is if this previously worked in some version and if the backend has changed such that it's not longer working in 42+? E.g. did api/system/tasks ever have GML_IMPORT as a prop with tasks? That might be something we want to ask the backend team about before merging? Or we can just merge and make a note to follow up with the backend team.

I'm approving because I'm not around next week, but I can also follow up more when I'm back if there's not pressure to merge

karolinelien:
@tomzemp you are correct - GML_IMPORT did once populate api/system/tasks, and the backend changed so it no longer does.

It changed in 2.41 (Aug 2023), in two back-to-back refactors:
[9e796fd748 — "feat: DB based job scheduling [DHIS2-15027]" (#14716)](https://github.com/dhis2/dhis2-core/pull/14716) deleted the setLocation("/system/tasks/" + GML_IMPORT) wiring and reduced GML_IMPORT to a bare, unused enum constant.
A week later, [0bb3934f72 — "async metadata import via scheduler import job [DHIS2-15702]" (#14912)](https://github.com/dhis2/dhis2-core/pull/14912) routed async GML (the application/xml branch of the shared MetadataImportJob) through the METADATA_IMPORT job type.

Current state:
Async GML import reports progress and its summary under METADATA_IMPORT - it shares MetadataImportJob, whose getJobType() returns METADATA_IMPORT.
GML_IMPORT still exists in the JobType enum but is dead code.

So for backend team, we could ask them to:
Either remove the orphaned GML_IMPORT JobType constant or give GML its own job type again so the app doesn't have to conflate GML progress with generic metadata imports. The current app workaround means a concurrent JSON/CSV metadata import and a GML import land in the same METADATA_IMPORT bucket. (But, is it worth the backend work is the question I guess.)

We can follow this up when you're back!

[DHIS2-15027]: https://dhis2.atlassian.net/browse/DHIS2-15027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-15702]: https://dhis2.atlassian.net/browse/DHIS2-15702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

tomzemp:
Since this app is compatible with v41+, this change will not need to be feature toggled to support v40 and earlier (I should have checked the minimum DHIS2 version before I added a feature toggle and then reverted it 😛)

We'll merge this to fix the issue and then follow up later with backend team to see if there's an objection to presenting GML as a separate job type within the app's UI.

PR #2251, opened 2026-07-27
fix: add alert when export request fails [DHIS2-18667]

Sharmyn28:
Add warning alert when using Export Event (CODE - idScheme)

[DHIS2-18667](https://dhis2.atlassian.net/browse/DHIS2-18667)

_Export event (CODE - idScheme)_
image

[DHIS2-18667]: https://dhis2.atlassian.net/browse/DHIS2-18667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/integration/EndpointParamChecks/ExportEvent.feature
+++ b/cypress/integration/EndpointParamChecks/ExportEvent.feature
@@ -84,3 +84,8 @@ Feature: The user should be able to export events
         Given the "inclusion" input is set to "CHILDREN"
         When the export form is submitted
         Then the download request is sent with the right parameters
+
+    Scenario: The export request fails
+        Given the event export request will fail
+        When the export form is submitted
+        Then a warning alert is shown with the error message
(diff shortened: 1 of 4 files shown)


src/utils/helper.js
```diff
@@ -194,12 +194,18 @@ const locationAssign = (url) => {
             const urlFilePart = new URL(downloadUrl).pathname.split('/').pop()
             const [, filename] = urlFilePart.match(/(^[^.]+)(\..+$)/)
```
tomzemp: this should be 
```
const [filename] = urlFilePart.match(/(^[^.]+)(\..+$)/)
```
From Claude, the bug existed before but wasn't noticeable because "the browser made an actual HTTP request and could infer the correct extension itself from the response's Content-Type/Content-Disposition headers, papering over the missing extension in download"

(there are a couple of tests that should be updated as well to include the file extension

tomzemp (commented):
Hi. Looks good in terms of reusing the request!

It's not working for me, however, to actually get the file contents (see my comment)

tomzemp (approved):

PR #2266, opened 2026-08-12
feat: add warning on import of trackedentities

Chisomchima:


src/pages/TEIImport/form-helper.js
```diff
@@ -1,6 +1,39 @@
...
+                'This file is not in a supported format. Files exported from DHIS2 2.40 or earlier are no longer supported by this app — please re-export the data using DHIS2 2.41 or later.'
+            ),
+        },
+    ],
+}
+
+const isLegacyTrackerPayload = async (file) => {
+    try {
+        const data = JSON.parse(await file.text())
+        return (
```
tomzemp: if this is a very specific check to prevent users from downloading in 2.40 earlier and uploading 2.41+, would it be sufficient to just check if the parsed object has `trackedEntityInstances` prop at root level because the format would be like `{trackedEntityInstances: [{...tracked entity details}]}`?

